PR #2, opened 2023-10-03
Telefonica API Spec initial contribution

fernandopradocabrillo:
*What is the purpose of this PR?*

Provide an initial API definition from Telefónica to start discussion

### DIFF
--- a/code/API_definitions/kyc-match.yaml
+++ b/code/API_definitions/kyc-match.yaml
@@ -1,0 +1,394 @@
+openapi: 3.0.3
+info:
+  title: Know Your Customer Match
+  description: |-
+    Allows to verify the matching of a number of attributes related to a customer identity against the verified account data bound to their phone number.
+
+    # Relevant Definitions and concepts
+
+    - **KYC**: stands for Know Your Customer and it is the process of a business verifying the identity of their clients and assessing their suitability, along with the potential risks of illegal intentions towards the business relationship.
+
+    # API Functionality
+
+    This API allows API clients to verify the matching of a number of attributes related to a customer identity against the account data bound to their phone number. The API is intended to be used in the following scenarios:
+
+    - To verify the user personal data during the digital registration of an account to a 3rd party service.
+    - To prevent fraud, wrong or imprecise information, and/or facilitate the onboarding of a user to a 3rd party service.
+
+    The API supports a multi-level hierarchy of property validation. In addition to the initial verification of the `phoneNumber`, an additional `idDocument` validation may occur based on different OB requirements. This means that if the value of `idDocument` is not provided or it does not match the one bound to the specific phone number in the OB systems, the operation will be return an error.
+
+    In order to increase the compatibility of the input data with those stored in the OB systems, some fields will be subject to a normalization process consisting mainly of the elimination of blank and special characters (such as scores, underscores or language-dependent characters) resulting in a concatenation of the different parts in cases where there are compound strings (such as including an additional lastName). Matching will be done in lower case by default unless there is a specific scenario since the normalization rules will be set by the OB and may vary from system to system.
+
+    **Note**: It may occur that the API response contains some values matched and some others not matched. In such cases, it is the responsibility of the the API Client to determine whether or not the identity verification is correct depending on the specific use case.
+
+    # Resources and Operations overview
+
+    The API provides the following endpoint:
+
+    - An endpoint to verify the matching of a number of attributes related to a customer identity against the account data bound to their phone number.
+
+    # Further info and support
+
+    (FAQs will be added in a later version of the documentation)    
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0-wip
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: kyc-match/0
+        description: Base path for the Know Your Customer API
+tags:
+  - name: Matching Identity
+    description: Operations to match a customer identity against the account data bound to their phone number.
+paths:
+  /match:
+    post:
+      tags:
+        - Matching Identity
+      summary: Matching a customer identity by checking a set of attributes related against the account data bound to their phone number.
+      description: |-
+        Verify matching of a number of attributes related to a customer identity against the verified account data bound to their phone number in the OB systems. Regardless of whether the `phoneNumber` is explicitly stated in the request body, at least one of the other fields must be provided, otherwise a `HTTP 400 - KNOW_YOUR_CUSTOMER.INVALID_PARAM_COMBINATION` error will be returned.
+
+        In order to proceed with the match check, there may be the requirement to perform a second level of validation based on the `idDocument` property. This means that in those cases, the `idDocument` is required and the provided value needs to match the one stored in the OB system associated with the indicated `phoneNumber`.
+        - If no `idDocument` is provided, then a `HTTP 403 - KNOW_YOUR_CUSTOMER.ID_DOCUMENT_REQUIRED` error will be returned.
+        - If the provided `idDocument` does not match the one stored in the OB systems, then a `HTTP 403 - KNOW_YOUR_CUSTOMER.ID_DOCUMENT_MISMATCH` error will be returned.
+
+        This operation is intended to be used with a **3-legged Access Token** so, if the phoneNumber is provided, it must match the `sub` in the Access Token; this is because only the user of the phone number must be able to request a verification of their identity.
+        If `phoneNumber` is not explicitly provided, the match will be done for the user based on the `sub` in the Access Token.
+
+        The API will return the result of the matching process for each requested attribute. This means that the response will **only** contain the attributes for which validation has been requested. Possible values are:
+          - **true**: the attribute provided matches with the one in the OB systems.
+          - **false**: the attribute provided does not match with the one in the OB systems.
+          - **not_available**: the attribute is not available to validate. **NOTE**: This value is only admissible when the information does not exist in the backend system or if it exists but it does not apply to the context of the phone number provided.
+      operationId: matchIdentity
+      security:
+        - openId:
+          - know-your-customer:match
+      parameters:
+        - $ref: '#/components/parameters/x-correlator'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MatchIdentityRequestBody'
+            examples:
+              MatchIdentityRequestBodyExample:
+                value:
+                  phoneNumber: '+34629255833'
+                  idDocument: 66666666q
+                  identity:
+                    firstName: Federica
+                    lastName: Sanchez Arjona
+                  address:
+                    postalCode: '47004'
+                    streetName: Nicolas Salmeron
+                    streetNumber: '4'
+                  birthdate: '1978-08-22'
+      responses:
+        '200':
+          description: OK
+          headers:
+            x-correlator:
+              $ref: '#/components/headers/x-correlator'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MatchIdentityResponse'
+              examples:
+                MatchIdentity200Example:
+                  value:
+                    idDocument: 'true'
+                    identity:
+                      firstName: 'true'
+                      lastName: 'true'
+                    address:
+                      postalCode: 'false'
+                      streetName: 'true'
+                      streetNumber: 'true'
+                    birthdate: not_available
+        '400':
+          $ref: '#/components/responses/KnowYourCustomer400InvalidArgument'
+        '401':
+          $ref: '#/components/responses/Generic401'
+        '403':
+          $ref: '#/components/responses/KnowYourCustomer403PermissionDenied'
+        '500':
+          $ref: '#/components/responses/Generic500'
+        '503':
+          $ref: '#/components/responses/Generic503'
+        '504':
+          $ref: '#/components/responses/Generic504'
+components:
+  securitySchemes:
+    openId:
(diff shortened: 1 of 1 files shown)


code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
+openapi: 3.0.3
+info:
+  title: Know Your Customer Match
+  description: |-
+    Allows to verify the matching of a number of attributes related to a customer identity against the verified account data bound to their phone number.
```
DT-DawidWroblewski, line 5: What is the purpose of this API? Is it to _verify the matching_ or verify user input vs CSP data that is behind MSISDN? This description assumes that CSP is _Identity Provider_ which might not be true.

what does it mean _verified account data_???

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+
+    # Relevant Definitions and concepts
+
+    - **KYC**: stands for Know Your Customer and it is the process of a business verifying the identity of their clients and assessing their suitability, along with the potential risks of illegal intentions towards the business relationship.
+
+    # API Functionality
+
+    This API allows API clients to verify the matching of a number of attributes related to a customer identity against the account data bound to their phone number. The API is intended to be used in the following scenarios:
```
DT-DawidWroblewski, line 13: Idem

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+
+    - To verify the user personal data during the digital registration of an account to a 3rd party service.
```
DT-DawidWroblewski, line 15: to verify user input of personal data

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+    - To prevent fraud, wrong or imprecise information, and/or facilitate the onboarding of a user to a 3rd party service.
+
+    The API supports a multi-level hierarchy of property validation. In addition to the initial verification of the `phoneNumber`, an additional `idDocument` validation may occur based on different OB requirements. This means that if the value of `idDocument` is not provided or it does not match the one bound to the specific phone number in the OB systems, the operation will be return an error.
```
DT-DawidWroblewski, line 18: I wouldn't add idDocument, because of lack of certainty about type of document used by MNOs.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+
+    In order to increase the compatibility of the input data with those stored in the OB systems, some fields will be subject to a normalization process consisting mainly of the elimination of blank and special characters (such as scores, underscores or language-dependent characters) resulting in a concatenation of the different parts in cases where there are compound strings (such as including an additional lastName). Matching will be done in lower case by default unless there is a specific scenario since the normalization rules will be set by the OB and may vary from system to system.
```
DT-DawidWroblewski, line 20: Is it assumed that this API only covers scenario with **attributes that are passed in plain text to MNO** ??? Please note that existing Mobile Connect KYC Match service, that is commercially available, covers **hashed** scenario, therefore data normalization is mandatory for developers/aggregators and CSPs.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+
+    **Note**: It may occur that the API response contains some values matched and some others not matched. In such cases, it is the responsibility of the the API Client to determine whether or not the identity verification is correct depending on the specific use case.
```
DT-DawidWroblewski, line 22: Again - this API description assumes that **Telco is an identity provider**, which might not be true in many countries.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+
+    # Resources and Operations overview
+
+    The API provides the following endpoint:
+
+    - An endpoint to verify the matching of a number of attributes related to a customer identity against the account data bound to their phone number.
+
+    # Further info and support
+
```
DT-DawidWroblewski, line 31: Reference to Mobile Connect KYC Match is expected here as well. [MC KYC spec](https://www.gsma.com/identity/wp-content/uploads/2023/01/IDY.23-v1.0.pdf)

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+paths:
+  /match:
+    post:
+      tags:
+        - Matching Identity
+      summary: Matching a customer identity by checking a set of attributes related against the account data bound to their phone number.
+      description: |-
+        Verify matching of a number of attributes related to a customer identity against the verified account data bound to their phone number in the OB systems. Regardless of whether the `phoneNumber` is explicitly stated in the request body, at least one of the other fields must be provided, otherwise a `HTTP 400 - KNOW_YOUR_CUSTOMER.INVALID_PARAM_COMBINATION` error will be returned.
+
+        In order to proceed with the match check, there may be the requirement to perform a second level of validation based on the `idDocument` property. This means that in those cases, the `idDocument` is required and the provided value needs to match the one stored in the OB system associated with the indicated `phoneNumber`.
```
DT-DawidWroblewski, line 62: Can you please explain _second level of validation_ concept?

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+        - If no `idDocument` is provided, then a `HTTP 403 - KNOW_YOUR_CUSTOMER.ID_DOCUMENT_REQUIRED` error will be returned.
+        - If the provided `idDocument` does not match the one stored in the OB systems, then a `HTTP 403 - KNOW_YOUR_CUSTOMER.ID_DOCUMENT_MISMATCH` error will be returned.
+
+        This operation is intended to be used with a **3-legged Access Token** so, if the phoneNumber is provided, it must match the `sub` in the Access Token; this is because only the user of the phone number must be able to request a verification of their identity.
```
DT-DawidWroblewski, line 66: MC KYC Match is also available with 2-legged token (client credentials). Please add this in description.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+        If `phoneNumber` is not explicitly provided, the match will be done for the user based on the `sub` in the Access Token.
+
+        The API will return the result of the matching process for each requested attribute. This means that the response will **only** contain the attributes for which validation has been requested. Possible values are:
+          - **true**: the attribute provided matches with the one in the OB systems.
+          - **false**: the attribute provided does not match with the one in the OB systems.
+          - **not_available**: the attribute is not available to validate. **NOTE**: This value is only admissible when the information does not exist in the backend system or if it exists but it does not apply to the context of the phone number provided.
```
DT-DawidWroblewski, line 72: what about a scenario, where data is available, but access denied?

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+              $ref: '#/components/schemas/MatchIdentityRequestBody'
+            examples:
+              MatchIdentityRequestBodyExample:
+                value:
+                  phoneNumber: '+34629255833'
+                  idDocument: 66666666q
+                  identity:
+                    firstName: Federica
+                    lastName: Sanchez Arjona
+                  address:
```
DT-DawidWroblewski, line 93: is it assumed that address has to be passed with separated fields? what about concatenated address field?

DT-DawidWroblewski (commented):
initial comments

ToshiWakayama-KDDI:
@fernandopradocabrillo, @DT-DawidWroblewski 

Thank you very much for the initial proposal and the initial comments.  

As the API onwer, I would suggest parking this YAML discussion for a while, because, 
- others may have their own proposals; actually we KDDI have our own proposal 
- this subproject maintainers (active contributors) who may have their proposals have not yet been on board so it may be better to wait for them
- as there are possibley more than one proposals, it is better to begin with discussions on use cases, flows, parameters and so on, before YAML discussion
- also we should not forget to consider alignment between Match and Fill-in

I will check the potential contributors and also set our regular meetings (we haven't) and start discussion on use cases/flows/parameters etc.

But your initial proposal to drive this API activity is really appreciated.

code/API_definitions/kyc-match.yaml
fernandopradocabrillo: In my understanding the purpose is to validate the input data against the data stored by the carrier related to the provided phone number. If the description is not accurate we can discuss updates no problem.
Yes, we asume that the owner of the customer information is the CSP

code/API_definitions/kyc-match.yaml
fernandopradocabrillo: The normalization rules can be different in every CSP/implementation so we propose that it can be done by the CSP applying the same rules to the input data and the stored data for the matching check

code/API_definitions/kyc-match.yaml
fernandopradocabrillo: We can include it, but why is it expected? what adds to this API an annotation indicating that there is a similar API with a different implementation?

code/API_definitions/kyc-match.yaml
fernandopradocabrillo: sure! in teh proposal the first thing we validate is the phone_number:

1. that it belongs to the CSP
2. if it is provided in the body request it should match the one in the authorization contest of the token
3. that the phone_number belongs to the user_id in sub
4. etc

then, there is an optional additional second level of validations with the field idDocument:

1. in Spain, for example, is mandatory to validate that the idDocument is provided along with the phone number
2. the idDocument provided should match with the one stored by the CSP and it should belong to the user
3. if no phone number is provided in the request, it should belong to the phone number in the access token

code/API_definitions/kyc-match.yaml
fernandopradocabrillo: Based on the recent agreement in the TSC meetings this API can only be used in 3-legged mode

code/API_definitions/kyc-match.yaml
StefanoFalsetto-CKHIOD: The address field has always been hard to manage. The concatenation rules are quite depending on country and it is not possible to easily create a "global" normalization algorithm. In our experience it is possible to have a high match rate only with separated fields. And even in that case it is needed to be crystal clear on what put into those fields. And there will always be exceptions.

code/API_definitions/kyc-match.yaml
StefanoFalsetto-CKHIOD: I think that this sentence it not preventing the Service Provider to perform the the normalization process before hashing the data sent to the MNO. Am I missing something?

Moreover, I propose to talk about the normalization algorithm. Maybe we can write something a bit more specific, to guide the developers/aggregators and CSPs to reach the same results, something like this:
For all the text-based attributes:
    1. Truncating string to a pre-defined maximum value, defined on a per-attribute base
    2. Lowercase the resulting string
    3. Remove special characters like punctuation and spaces (maybe we can use the definition of [:punct:] and [:space:] classes of POSIX standard)
    4. Degrade all the characters to the underlying Latin-1 (e.g., À --> A, Š -> S etc.)

code/API_definitions/kyc-match.yaml
StefanoFalsetto-CKHIOD: @fernandopradocabrillo in your last comment you state that the purpose of KYC Match is to: 

> validate the input data against the data stored by the carrier related to the provided phone number

And I agree on that. I am also interested in exactly defining the "validate" term. Is it just as a string comparison?

From our experience, the match rate obtained using exact string match is most of the time low. And that's a problem coming from non-standard way to collect data, not only from CSPs but sometimes also from carriers.
Here we are defining a new service, hence we can try to maximize the match score, improving the specifications.
  
Hence I am proposing to expand the "match" concept introducing the "string similarity" metric. This could respond to current revenue streams (i.e., similarity=1, full match) and also create newer ones (i.e., similarity<1, partial match).

What do you think? In case this new metric is of interest, we can talk about the way to evaluate it.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+  /match:
+    post:
+      tags:
+        - Matching Identity
+      summary: Matching a customer identity by checking a set of attributes related against the account data bound to their phone number.
+      description: |-
+        Verify matching of a number of attributes related to a customer identity against the verified account data bound to their phone number in the OB systems. Regardless of whether the `phoneNumber` is explicitly stated in the request body, at least one of the other fields must be provided, otherwise a `HTTP 400 - KNOW_YOUR_CUSTOMER.INVALID_PARAM_COMBINATION` error will be returned.
+
+        In order to proceed with the match check, there may be the requirement to perform a second level of validation based on the `idDocument` property. This means that in those cases, the `idDocument` is required and the provided value needs to match the one stored in the OB system associated with the indicated `phoneNumber`.
+        - If no `idDocument` is provided, then a `HTTP 403 - KNOW_YOUR_CUSTOMER.ID_DOCUMENT_REQUIRED` error will be returned.
```
StefanoFalsetto-CKHIOD, line 63: Maybe we can create a more generic "second level of validation". 
In that case this is an error specific for a single use case for a single MNO in a single country.

code/API_definitions/kyc-match.yaml
```diff
@@ -0,0 +1,394 @@
...
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        idDocument:
+          $ref: '#/components/schemas/IdDocument'
+        identity:
+          $ref: '#/components/schemas/Identity'
+        address:
+          $ref: '#/components/schemas/Address'
+        birthdate:
+          type: string
```
StefanoFalsetto-CKHIOD, line 156: Maybe we can use some pattern to explicitly state what we are expecting:
```
                pattern: "^\\d{4}-\\d{2}-\\d{2}$"
```

odiego:
Hello folks, any updates in approving this initial spec?

fernandopradocabrillo:
Closing as we already have an initial proposal in the repo

PR #14, opened 2023-11-13
KDDI API KYC-match Initial Contribution

ToshiWakayama-KDDI:
#### What type of PR is this?

* enhancement/feature

* Provide an initial API definition from KDDI for discussion on KYC Match YAML.
 
NA

* The YAML is almost same as the one currently provided by KDDI as commercial services.  As our original YAML has descriptions/remarks/notes written in Japanese, those have been translated into English.
 
Also, please note that it has some attributes/parameters that seem specific to Japanease language.  It is intentional, because these attributes/parameters are needed by our customers in Japan and we would like to discuss how to handle this kind of country/regional requirements in API specifications.  

#### Changelog input

```
 release-note

```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,4 +6,4 @@
 
 # These are the default owners for the whole content of this repository. The default owners are automatically added as reviewers when you open a pull request, unless different owners are specified in the file.
 
-* @ToshiWakayama-KDDI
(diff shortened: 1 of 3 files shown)


ToshiWakayama-KDDI:
Something wrong.  Let me cancel this PR.

PR #27, opened 2023-12-12
Consolidated KYC Match API definition

ToshiWakayama-KDDI:
#### What type of PR is this?

* enhancement/feature

#### What this PR does / why we need it:

* Provide a consolidated KYC Match API definition based on subproject discussion for the initial version.

PR #28, opened 2023-12-12
Consolidated KYC Fill-in API definition

ToshiWakayama-KDDI:
#### What type of PR is this?

* enhancement/feature

* Provide a consolidated KYC Fill-in API definition based on subproject discussion for the initial version.

### DIFF
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -50,7 +50,7 @@ paths:
             examples:
               KYC_FillinRequestBodyExample:
                 value:
-                  cp_id: 12345
+                  3rdPartyId: 12345
 
       responses:
         '200':
@@ -63,26 +63,26 @@ paths:
               examples:
                 KYC_Fillin200Example:
                   value:
-                    phoneNumberFillin: '+34629255833'
-                    idDocumentFillin: 66666666q
-                    nameFillin: Federica Sanchez Arjona
-                    giventNameFillin: Federica
-                    familyNameFillin: Sanchez Arjona
-                    nameKanaHankakuFillin: federica
-                    nameKanaZenkakuFillin: Ｆｅｄｅｒｉｃａ
-                    middleNamesFillin: Sanchez
-                    familyNameAtBirthFillin: YYYY
-                    addressFillin: Tokyo-to Chiyoda-ku Iidabashi 3-10-10
-                    streetNameFillin: Nicolas Salmeron
-                    streetNumberFillin: 4
-                    postalCodeFillin: 1028460
-                    regionFillin: Tokyo
-                    localityFillin: ZZZZ
-                    countryFillin: Japan
-                    houseNumberExtensionFillin: VVVV
-                    birthdateFillin: '1978-08-22'
-                    emailFillin: abc@example.com
-                    genderFillin: male
+                    phoneNumber: '+34629255833'
+                    idDocument: 66666666q
+                    name: Federica Sanchez Arjona
+                    giventName: Federica
+                    familyName: Sanchez Arjona
+                    nameKanaHankaku: federica
+                    nameKanaZenkaku: Ｆｅｄｅｒｉｃａ
+                    middleNames: Sanchez
+                    familyNameAtBirth: YYYY
+                    address: Tokyo-to Chiyoda-ku Iidabashi 3-10-10
+                    streetName: Nicolas Salmeron
+                    streetNumber: 4
+                    postalCode: 1028460
+                    region: Tokyo
+                    locality: ZZZZ
+                    country: Japan
+                    houseNumberExtension: 36
+                    birthdate: '1978-08-22'
+                    email: abc@example.com
+                    gender: male
 
         '400':
           $ref: '#/components/responses/Generic400'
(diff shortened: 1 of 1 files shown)


GillesInnov35:
Hi all, I think we should not put a suffix xxxxFillin to response attribute because the API Fill-In returns same attributes that the Match request has.
So to my opinion we should have the same attribute naming.

ToshiWakayama-KDDI:
Hi @GillesInnov35 , Hi All,

Understand.  I do not have a strong view, but I think a Match response is not always 'true'.  For example, if a Match request [name:'Steven Chen'] is false and the OB system has 'Stephen Chen', a Fill-in response will be [name:'Stephen Chen'].  So, these two 'name's are different.  
Considering this kind of case, it may be better to have different suffixes for Match request attributes and Fill-in response attributes to differentiate a Match request (sometimes false) and a Fill-in response (always correct).

What do you and others think?

Many thanks,
Toshi

ToshiWakayama-KDDI:
Hi @GillesInnov35 , Hi All,

In addtion to my previous comment, what about Fill-in request?

I am sharing one point here.  As you can see in the Consollidated YAML, currently there is only one in the Fill-in request body.  The reason is that our (KDDI's) Fill-in service is as follows: First, 3rd parties/API consumers make a service contract with API service provider (KDDI), when they receive 3rd party ID and they have to decide what parameters/attributes they will receive. Then, when using the KYC Fill-in API, they will send their 3rd party ID only in the request body, and then API service provider (KDDI) will return parameters/attributes the 3rd party/API consumer has decided by the contract.  One of the reasons is from security/privacy point of view.

This will be written down in the description.

Many thanks,
Toshi

GillesInnov35:
hi @ToshiWakayama-KDDI , I think suffix fillin should not be added to the attributes name. The fill-In API response returns same **value type** of attribute as the match request even if the value could be different. This is two different things.

That's why I think the name of attribute should be the same without any fillin suffix. If we add multiple attribute names for the same field object it will be confused for consumer.
If we take the example of Match API, in Match request and Match response and explain why the names are different. The value type of attribute are different. For example **_givenName_** in the request is the given name of the person and _**givenNameMatchResult**_ is a result of a comparison. 

To all, what is your point of view ?
BR
Gilles

ToshiWakayama-KDDI:
Hi  @GillesInnov35 ,

Thank you for your response.  Understood.  In terms of '**value type**', you are correct.  No Fillin suffix is fine with us.

Best regareds,
Toshi

ToshiWakayama-KDDI:
> Hello Colleagues, We strongly believe that the KYC Fill-in Request should also include the users MSISDN or a respective Pseudonymous Identifier. @ToshiWakayama-KDDI I do not understand how a response with a specific customer's KYC parameters can be provided if there is no identifier for the specific customer in the respective request.

Hi @EfthymisIsaakidis-DTCS ,
Thank you for your comment on the KYC Fill-in Request body.  Actually we do not need the end user identity (phone number) in the request body, because it is captured during AuthN/ AuthZ/ Consent capturing process and it is contained in the Token.  This needs to be documented somewhere in the API document/YAML, I understand.

Best regards,
Toshi

GillesInnov35:
@ToshiWakayama-KDDI, could you please explain why you think the mobile phone number should not be a request field.
Phone number is part of most of the CAMARA API request attributes, It is the case of Match request, Number Verify, device status, Device identifier, Sim Swap, etc. 
Thanks a lot
Gilles

GillesInnov35:
I had a discussion with my colleague L. Robert who is working on few CAMARA APIs design. He confirms that even if 3-Legged Authent allow retreiving the msisdn, phone number field has been added to most of API definition to allow 2-Legged Authentication offer and tests from application server.
So we have to decide if we keep it in the request definition as it is in atch request or not.
BR
Gilles

ToshiWakayama-KDDI:
Hi ＠Gilles, 
Thanks for your comment.  So, I understand that 3-Legged is necessary for KYC Fill-in as it handles customer's privacy information, so, we could say that Phone Number is not neccessary.  But I feel that if phone number is needed, it can be added as an optional parameter.  I need to discuss with my colleagues.
Best regards,

ToshiWakayama-KDDI:
Hi @GillesInnov35 , Hi @fernandopradocabrillo ,

I have made all the required corrections, in alignment with Match, including no attribute Request Body to the Conslidated Fill-in YAML.  I have just checked by Swagger and got no errors.  Please approve this PR, so that I can merge it.

Many thanks,
Toshi

ToshiWakayama-KDDI:
Hi @GillesInnov35 , HI @fernandopradocabrillo ,

Could you please approve this PR.

Thanks,
Toshi

ToshiWakayama-KDDI:
Hi @GillesInnov35 , HI @fernandopradocabrillo ,

Could you please approve this PR.

Thanks,
Toshi

ToshiWakayama-KDDI:
Hi @fernandopradocabrillo , Hi @GillesInnov35 .

I have just merged the KYC Match PR #27.  Thank you so much for your cooperation.

If you could approve this Fill-in PR without delay, it would be much appreciated.

Many thanks,
Toshi

GillesInnov35 (approved):

GillesInnov35:
Done

ToshiWakayama-KDDI:
Thank you very much!  Toshi

PR #60, opened 2024-03-08
chore/remove-phoneNumberMatch

### LINKED ISSUE #59: Remove phoneNumberMAtch field from response schema
**Problem description**
The proposal is to remove the property `phoneNumberMatch` from the response schema since we believe it is not necessary. What is it going to match against? if the phoneNumber provided doesn't match the one in the token the API will return a 403. And the client should not be able to send a phone number different from the one that the user is authenticated with, it will always return 403.

So `phoneNumberMatch` will always return true since the phone number will always have to match otherwise there won't be a 200.

fernandopradocabrillo:
#### What type of PR is this?

* correction

Remove phoneNumberMatch from KYC match response

Fixes #59

### DIFF
--- a/code/API_definitions/kyc-match.yaml
+++ b/code/API_definitions/kyc-match.yaml
@@ -177,7 +177,7 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
-   
+
   parameters:
     x-correlator:
       name: x-correlator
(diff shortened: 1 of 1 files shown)


GillesInnov35 (dismissed):
OK thanks

fernandopradocabrillo:
> OK thanks

@GillesInnov35 can you please review again? I forgot to remove the field from the example 😄

GillesInnov35 (approved):
OK

ToshiWakayama-KDDI:
Hi @fernandopradocabrillo,

Thanks very much.  LGTM.

BR
Toshi

ToshiWakayama-KDDI (approved):

PR #137, opened 2024-08-20
Remove countryMatchScore property

### LINKED ISSUE #134: KYC Match : CountryMatchScore makes no sense
In the current list of attributes for which a MatchScore can be supplied in case the Match is false, also Country is included.

Country is however now become formatted as ISO 3166-1 alpha-2, so I don't think it makes much sense here to supply a MatchScore anymore

fernandopradocabrillo:
#### What type of PR is this?

* correction

Since Country property is now in ISO 3166 alpha-2 codification, so it doesn't make sense to have an score for this property.

Fixes #134 

yaml version has been set to wip in case we have to include more changes before the final release. At that moment, the release PR will contain the final version change.

### DIFF
--- a/code/API_definitions/kyc-match.yaml
+++ b/code/API_definitions/kyc-match.yaml
@@ -49,7 +49,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.0-rc.1
+  version: wip
 
   x-camara-commonalities: 0.4.0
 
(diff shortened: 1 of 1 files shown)


GillesInnov35 (approved):
Thanks Fernando

GillesInnov35:
@fernandopradocabrillo , @ToshiWakayama-KDDI , I wonder if we should not apply the same rules for streetNumber and remove streetNameMatchScore as it will be done for countryCode and postalCode.
What do you think about that ?
BR
Gilles

ToshiWakayama-KDDI:
Hi @GillesInnov35 , cc @fernandopradocabrillo 

Sorry for the delay.  LGTM.  So I will approve this, and, as I understand Gilles's comment above has been cleared in Issue #134, I will merge this PR.

BR
Toshi

ToshiWakayama-KDDI (approved):

PR #148, opened 2024-09-10
fix type and change version

### LINKED ISSUE #112: [KYC-Match] MatchScoreResult type integer or float
**Problem description**
Currently the `MatchScoreResult` is defined as float but have the ranges 0-100 typical for percentage values.

```
    MatchScoreResult:
      type: number
      format: float 
      description: Indicates the similarity score assigned to the input value when it does not exactly match the value stored in the operator's system. This property shall only be returned when the value of the corresponding match field is `false`.
      minimum: 0
      maximum: 100
```

**Expected behavior**

Add description the value is percentage.

If more precision is not needed the `MatchScoreResult`  can be defined as as integer :

```
    MatchScoreResult:
      type: integer
```

**Alternative solution**
Keep  `MatchScoreResult` defined as float and change the range to 0-1 adding in the description that it should be limited to 2 or 3 decimal places.
```
      minimum: 0
      maximum: 1
```

**Additional context**
In Location Verification "MatchRate" is defined as integer.
https://github.com/camaraproject/DeviceLocation/blob/main/code/API_definitions/location-verification.yaml#L388C8-L388C15

fernandopradocabrillo:
#### What type of PR is this?

* bug
* correction

Fixes typo in score property type

Fixes #112 

This PR changes the version back to wip. With the creation of the new release it will be 0.2.1

### DIFF
--- a/code/API_definitions/kyc-match.yaml
+++ b/code/API_definitions/kyc-match.yaml
@@ -51,7 +51,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.0
+  version: wip
 
   x-camara-commonalities: 0.4.0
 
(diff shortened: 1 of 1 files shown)


GillesInnov35 (approved):
LGTM  thanks

PR #153, opened 2024-10-15
KYC-Fill-In: Use the same values in the description and example - lowercase

AxelNennker:
#### What type of PR is this?

* documentation

The description uses different values than the example. This PR changes the description.

There is no issue because this is only just removing an inconsistency


GillesInnov35 (approved):
Thanks

fernandopradocabrillo (approved):

ToshiWakayama-KDDI (approved):
Thanks.

PR #155, opened 2024-10-17
KYC-Fill-In: define scopes

### LINKED ISSUE #90: [KYC Fill-in] Attributes to be included in the request body?
**Problem description**

For KYC Fill-in, in addition to the current 'none in the request body', to consider including Phone Number?, 3rdPartyId?, attributes which the 3rdParty wants to receive.
(Spin off from Issue #65, item No.10, as per Action Item #13.03)

BR

AxelNennker:
The current KYC Fill-In API does not fulfill the [GDPR principle of data minimization.](https://gdpr-info.eu/art-5-gdpr/#:~:text=adequate%2C%20relevant%20and%20limited%20to%20what%20is%20necessary%20in%20relation%20to%20the%20purposes%20for%20which%20they%20are%20processed%20(%E2%80%98data%20minimisation%E2%80%99))

#### What type of PR is this?

* correction

This PR defines the scopes that the API consumer can use when sending an [OIDC authentication request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) or an [CIBA authentication request](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1).

Removed 2024-10-29: references to standard claims defined in OIDC. I think it might be a separate discussion why CAMARA KYC uses e.g. givenName and not given_name. As defined in OIDC standard claims.

Fixes #90 

OpenAPI specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#security-requirement-object

> If the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names required for the execution

Which means that the API Consumer can request using `scope="openid kyc-fill-in:fill-in kyc-fill-in:phoneNumber name".
The access token would then have the scopes associated with it and the API producer will then find this association and the security requirement would be fulfilled. Then the API producer can go through the list of associated scopes and can add the requested attributes to the KYC Fill-In response.

### DIFF
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -79,8 +79,46 @@ paths:
 
       security:
         - openId:
-          - kyc-fill-in:fill-in
-
+          - kyc-fill-in:fill-in kyc-fill-in:phoneNumber
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:idDocument
+        - openId:
+          - kyc-fill-in:fill-in name
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:givenName
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:familyName
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:nameKanaHankaku
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:nameKanaZenkaku
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:middleNames
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:familyNameAtBirth
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:address
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:streetName
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:streetNumber
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:postalCode
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:region
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:locality
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:country
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:houseNumberExtension
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:birthdate
+        - openId:
+          - kyc-fill-in:fill-in email
+        - openId:
+          - kyc-fill-in:fill-in kyc-fill-in:gender
+        
       parameters:
         - $ref: '#/components/parameters/x-correlator'
 
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -79,45 +79,45 @@ paths:
 
       security:
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:phoneNumber
+          - kyc-fill-in:phoneNumber
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:idDocument
+          - kyc-fill-in:idDocument
         - openId:
-          - kyc-fill-in:fill-in name
+          - kyc-fill-in:name
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:givenName
+          - kyc-fill-in:givenName
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:familyName
+          - kyc-fill-in:familyName
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:nameKanaHankaku
+          - kyc-fill-in:nameKanaHankaku
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:nameKanaZenkaku
+          - kyc-fill-in:nameKanaZenkaku
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:middleNames
+          - kyc-fill-in:middleNames
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:familyNameAtBirth
+          - kyc-fill-in:familyNameAtBirth
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:address
+          - kyc-fill-in:address
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:streetName
+          - kyc-fill-in:streetName
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:streetNumber
+          - kyc-fill-in:streetNumber
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:postalCode
+          - kyc-fill-in:postalCode
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:region
+          - kyc-fill-in:region
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:locality
+          - kyc-fill-in:locality
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:country
+          - kyc-fill-in:country
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:houseNumberExtension
+          - kyc-fill-in:houseNumberExtension
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:birthdate
+          - kyc-fill-in:birthdate
         - openId:
           - kyc-fill-in:fill-in email
         - openId:
-          - kyc-fill-in:fill-in kyc-fill-in:gender
+          - kyc-fill-in:gender
         
       parameters:
         - $ref: '#/components/parameters/x-correlator'
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -79,6 +79,8 @@ paths:
 
       security:
         - openId:
+          - kyc-fill-in:all
+        - openId:
           - kyc-fill-in:phoneNumber
         - openId:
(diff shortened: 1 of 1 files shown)


GillesInnov35:
Thanks @AxelNennker, @HuubAppelboom, I totally agree that data minimization must be addressed following GDPR recommendations. That was a subject we had to work. But I'm quite surprised we should answer by multiple scopes per attribute.
Could we try to find another way to address this need or is it also a rules ?
BR
Gilles

GillesInnov35:
if we see what TMF recommends using fields syntax in the query to do something similar we should have use a GET operation.

HuubAppelboom:
@AxelNennker @GillesInnov35 
There are also some alternatives, for example using "profile" in the scope to return all name related items
See https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

Another option is to use claims as an Authorization request parameter 
See https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
That has the advantage that you can also specify in user consent cases which claims are essential for the API Consumer and which ones are voluntary (optional).  This is something that can very well be used in the consent acquisition process as well.

AxelNennker:
@GillesInnov35 

Maybe you are wondering about the kyc-fill-in:fill-in scope value? I started with that because I wanted to use the OIDC [standard claim](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) in this API. Then I discovered that this API does not use the OIDC standard claims. There are minimal differences sometimes e.g. given_name vs givenName.

I am going to remove the scope value kyc-fill-in:fill-in from this PR and always use the the attributes as scope.

The scope value in the authorization request would be "kyc-fill-in:givenName kyc-fill-in:familyName" if the API Consumer wants an access token hat allows the givenName and the familyName to be retrieved in the API call. The API call would not have any attributes as parameters because givenName and familyName would be associated with the access token.

AxelNennker:
I closed the discussion https://github.com/camaraproject/KnowYourCustomer/discussions/158
It is not so important for this PR.

This a minimal PR that enables data-minimization.
I think it is not good that the API consumer calls this API and the API provider provides whatever they have in their subscriber database.

As for the "alternatives", yes "[claims](https://openid.net/specs/openid-connect-core-1_0.html#Claims)" could be an alternative. Although, I think, the "essential" claims and optional claims also gets the API provider in hot GDPR water.
E.g. when Telekom Deutschland electronically reads attributes from the German id card, then our DPO advised us to read exactly read the attributes that we are required by law ([TKG §172 in German](https://www.gesetze-im-internet.de/tkg_2021/__172.html)) to collect - not more.
But that is another discussion, I think.

ToshiWakayama-KDDI:
Hi @AxelNennker , Hi all, 

Ater yesterday's meeting, I have discussed this internally with my colleagues.  Actually, what we are doing for the KYC Fill-in service is that, when we make contract with RPs for the KYC Fill-in API use, it is decided which parameters/attributes each RP (App client) will receive by the API call, based on each RP's requirement.  Each RP's requirement for KYC Fill-in API is usually stable, and by this process, we conform to data-minimization concept.

So, our view is that for non-GDPR countries including Japan, the proposed 'atomized scope' may be too much, a kind of bother to API users, potentially creating some side effects e.g. it may result in a extremely long URL because of many scopes with percent-encoding.  So we tend to be negative.  But, at the same time, GDPR countries' circumstances are understandable.

One proposal.  **Why don't we keep 'kyc-fill-in:fill-in' and just add atomized scope proposed by this PR, to fulfill both sides' requirements?**

Thanks,
Toshi

AxelNennker:
@ToshiWakayama-KDDI 
> Ater yesterday's meeting, I have discussed this internally with my colleagues. Actually, what we are doing for the KYC Fill-in service is that, when we make contract with RPs for the KYC Fill-in API use, it is decided which parameters/attributes each RP (App client) will receive by the API call, based on each RP's requirement. Each RP's requirement for KYC Fill-in API is usually stable, and by this process, we conform to data-minimization concept.

As you know, the agreement on allowed scope is done in CAMARA during the onboarding process of the API consumer. 
There is no direct contact between API producer and API consumer.
Detemining which scopes are allowed for an API consumer is the standard CAMARA onboarding process.
If KDDI wants to offer the API directly to API consumers, that is of course still possible, but I think that hurts the "spirit" of CAMARA. Together we are stronger.

> So, our view is that for non-GDPR countries including Japan, the proposed 'atomized scope' may be too much, a kind of bother to API users, potentially creating some side effects e.g. it may result in a extremely long URL because of many scopes with percent-encoding. So we tend to be negative. But, at the same time, GDPR countries' circumstances are understandable.

Current URL limits are at about 4k, I think, which should fit even long lists of scopes.
Put the API consumer can always POST the scopes in the OIDC authentication request which puts the parameters into the request body and URL length is not a problem. 

Please see: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
> Authorization Servers MUST support the use of the HTTP GET and POST methods defined in [RFC 7231](https://openid.net/specs/openid-connect-core-1_0.html#RFC7231) [RFC7231] at the Authorization Endpoint.

> One proposal. **Why don't we keep 'kyc-fill-in:fill-in' and just add atomized scope proposed by this PR, to fulfill both sides' requirements?**

OIDC also defines some scopes like "profile" that means that a collection of attributes is returned in the API response.
We could certainly define some special scopes that make sense for API consumers e.g. "kyc-fill-in:profile"
I would please request that those special scopes are well defined.
In my opinion a special scope like 'kyc-fill-in:fill-in' should not mean "I give you e.g. idDocument if I have it, but maybe not".
There could be a special scope address that means all attributes related to addresses. Also, please well defined.

Merge this PR and then create a new one that defines the special scopes?

//Axel

ps: I am not too happy that this API returns each attribute that is part of an address separately. 
Both OIDC and [eKYC-Identity Assurance](https://openid.net/wg/ekyc-ida/specifications/) have a scope "address" that returns the address as a JSON object.
It is then also easier to have scopes for billing address and fixed-line-address in the future.
See OIDC standard claims
> End-User's preferred postal address. The value of the address member is a JSON [[RFC8259]](https://openid.net/specs/openid-connect-core-1_0.html#RFC8259) structure containing some or all of the members defined in [Section 5.1.1](https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim).
Changing that would be another PR...

Macke:
Hi Camara people, Just a couple of thoughts for you to add into your decision making...
The PAR spec https://datatracker.ietf.org/doc/html/rfc9126 solves a few issues with the authorization endpoint.  You should really consider it for multiple reasons - 1 it removes the browser url size issue and it means that the request parameters do not need to pass via the browser.
I also think that the claims request parameter is pretty useful as it may allow the data consumer to request less than they are technically allowed (so even better from a GDPR perspective) without affecting the possibility of the provider having policies that restrict what they will share with any given consumer.
One potential issue with the claims request parameter is that there are only elements within it that cater for id_token or userinfo endpoints but it is readily extended if you were to wish to use it in relation to your kyc-fill-in API.

ToshiWakayama-KDDI:
Thanks, @AxelNennker , @Macke ,

We do not have a strong opinion on which to be used, scopes or claims parameters, for each kyc-fill-in parameters, first of all.  

From KDDI point of view, considering our API Customers' needs, we would still prefer to use scopes including a few or several parameters.  For data minimization, taking Axel's suggestion into account, we could **define special scopes, each of which would have a specific set of parameter**s.  

So the proposal is to **add a special scope which includes all parameters** to this PR #155, and then to define additional special scopes in later releases.  **The proposed special scope including all parameters is: 'kyc-fill-in:set-all'.**  It is easily understood that the scope is for all the parameters.  (Meaning easier than 'kyc-fill-in:fill-in').

Best regards,
Toshi

AxelNennker:
> So the proposal is to **add a special scope which includes all parameters** to this PR #155, and then to define additional special scopes in later releases. **The proposed special scope including all parameters is: 'kyc-fill-in:set-all'.** It is easily understood that the scope is for all the parameters. (Meaning easier than 'kyc-fill-in:fill-in').

So 'kyc-fill-in:set-all' would mean this list:

kyc-fill-in:phoneNumber
kyc-fill-in:idDocument
kyc-fill-in:name
kyc-fill-in:givenName
kyc-fill-in:familyName
kyc-fill-in:nameKanaHankaku
kyc-fill-in:nameKanaZenkaku
kyc-fill-in:middleNames
kyc-fill-in:familyNameAtBirth
kyc-fill-in:address
kyc-fill-in:streetName
kyc-fill-in:streetNumber
kyc-fill-in:postalCode
kyc-fill-in:region
kyc-fill-in:locality
kyc-fill-in:country
kyc-fill-in:houseNumberExtension
kyc-fill-in:birthdate
kyc-fill-in:email
kyc-fill-in:gender

If the API Provider does not know the attribute then that value would not returned.

Taking the example from kyc-fill-in the response might be:

```
                    phoneNumber: '+34629255833'
                    idDocument: 66666666q
                    name: Federica Sanchez Arjona
                    givenName: Federica
                    familyName: Sanchez Arjona
                    nameKanaHankaku: federica
                    nameKanaZenkaku: Ｆｅｄｅｒｉｃａ
                    middleNames: Sanchez
                    familyNameAtBirth: YYYY
                    address: Tokyo-to Chiyoda-ku Iidabashi 3-10-10
                    streetName: Nicolas Salmeron
                    streetNumber: 4
                    postalCode: 1028460
                    region: Tokyo
                    locality: ZZZZ
                    country: JP
                    houseNumberExtension: 36
                    birthdate: '1978-08-22'
                    email: abc@example.com
                    gender: male
```

If the API Provider would not know e.g. idDocument or e.g. nameKanaHankaku then that value would be missing.

ToshiWakayama-KDDI:
Hi @AxelNennker ,

Thanks very much for your proposal 'add section defining kyc-fill-in:set:all'.  It is fine with us.
Hence, the PR is fine with us.

Thanks,
Toshi
KDDI

ToshiWakayama-KDDI (approved):

GillesInnov35 (approved):
LGTM  Thanks

fernandopradocabrillo (approved):

Issue #161, opened 2024-10-30
Error in the name of the attribute middleNamesScore


GillesInnov35:
**Problem description**
There's an error in the name of the attribute _middleNamesScore_ - not consistent with the others, **_Match_** is missing

**Expected behavior**
to be consistent with the oters attributes name it must be: _middleNames**Match**Score_

fernandopradocabrillo:
Thanks for pointing it out Gilles. This change is tricky because even it is clearly an error, fixing it breaks the compatibility, specially if the API is already being integrated, which is our case. We can discuss what could be the best approach, but I think it should be at least in a v0.3..

GillesInnov35:
hi @fernandopradocabrillo , yes you're right there'd be a problem of backward compatibility with the correction. At Orange France we don't currently manage middle names but Spain is concerned. As you propose we can discuss in this issue about decision to be made as quicky as possible; it could be to keep the current name with the error and wait for the next version.
Thanks
BR
Gilles

fernandopradocabrillo:
For us the solution of leaving it like it is right now, at least for version 0.2.1 is alright. If the rest is ok with the decission, we can include it in higher versions (i.e v0.3)

ToshiWakayama-KDDI:
I understand PR #171 has solved this issue, so, can we close this issue?

PR #171, opened 2024-12-11
Correct name of middleNames Match Score attribute in kyc-match yaml file

GillesInnov35:
Correct name of middleNames Match Score attribute

#### What type of PR is this?

* correction

Correct name of middleNames Match Score response attribute. 
Replace middleNamesScore by **middleNamesMatchScore**

Fixes #https://github.com/camaraproject/KnowYourCustomer/issues/161

#### Changelog input

```
 release-note

```

### DIFF
--- a/code/API_definitions/kyc-match.yaml
+++ b/code/API_definitions/kyc-match.yaml
@@ -349,7 +349,7 @@ components:
           allOf:
             - $ref: '#/components/schemas/MatchResult'
             - description: Indicates whether the middle names of the customer matches with the one on the Operator's system.
-        middleNamesScore:
+        middleNamesMatchScore:
           $ref: '#/components/schemas/MatchScoreResult'
         familyNameAtBirthMatch:
           allOf:


HuubAppelboom:
> hi all, this update will break backward compatibility. So could be implemented only in a major version. WDYT BR Gilles

For KPN no problem, we have not gone live yet with customers

fernandopradocabrillo:
Thanks @GillesInnov35 for the PR. We are actually integrating clients with current version, so let me check the impact with my product team.

GillesInnov35:
thanks  @fernandopradocabrillo, yes there's an impact for current implementation using this attribute. It is not the case at Orange France but it could for others countries. So let's see impact to decide then if we delay this evolution to a major version I think.
BR
Gilles

GillesInnov35:
hi @all, do you think we'll correct this mistake in a future version? Thanks 
Gilles

HuubAppelboom:
> hi @ALL, do you think we'll correct this mistake in a future version? Thanks Gilles

For KPN it is no issue to fix it now, we have not gone live yet with customers

ToshiWakayama-KDDI:
> hi @ALL, do you think we'll correct this mistake in a future version? Thanks Gilles

Agree that sooner or later this will need to be fixed.

For KDDI, it is no issue to fix it now, since we have not implemented the Scoring function yet.

Thanks,
Toshi

hdamker:
> So let's see impact to decide then if we delay this evolution to a major version I think.

@GillesInnov35 @fernandopradocabrillo please consider that you are still in the phase of initial versions with 0.x. That means that there is no assumption of backward compatibility between 0.x and 0.x+1 (cf. https://semver.org/#spec-item-4). That's the reason that we are distinct the versions for 0.x also within the server url with .../v0.x 

Therefore I would say, that now is good point in time to introduce such fixes.

GillesInnov35:
hi @ToshiWakayama-KDDI , could you approve the PR if OK for you of course. 
Thanks a lot
BR
Gilles

ToshiWakayama-KDDI (approved):
Thanks a lot.  Sorry for the delay.

GillesInnov35:
Thanks a lot @ToshiWakayama-KDDI

PR #174, opened 2024-12-13
Add PhoneNumber In FillIn Request Body

GillesInnov35:
Add phoneNumber In Fill-in Request Body

#### What type of PR is this?

* enhancement/feature

Add phoneNumber attribute In kyc-fill-in Request Body as optionnal

Fixes #https://github.com/camaraproject/KnowYourCustomer/issues/90

### DIFF
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -93,7 +93,8 @@ paths:
             examples:
               KYC_FillInBodyExample:
                 value:
-                  phoneNumber: '+34629255833'
+                  { "phoneNumber": "+34629255833" }
+
       
       responses:
         '200':
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -84,6 +84,17 @@ paths:
       parameters:
         - $ref: '#/components/parameters/x-correlator'
 
+      requestBody:
+        content:
+          application/json:
+
+            schema:
+              $ref: '#/components/schemas/KYC_FillinRequest'
+            examples:
+              KYC_FillInBodyExample:
+                value:
+                  phoneNumber: '+34629255833'
+      
       responses:
         '200':
           description: OK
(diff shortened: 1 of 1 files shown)


ToshiWakayama-KDDI:
Thank you, @GillesInnov35 .  LGTM. 

Best regards,
Toshi
KDDI

AxelNennker (changes requested):
Is it customary to have an example both in the parameter description and in the schema definition? https://github.com/camaraproject/NumberVerification/blob/main/code/API_definitions/number-verification.yaml#L93 does have an example in the schema definition but there is no request example.

code/API_definitions/kyc-fill-in.yaml
```diff
@@ -84,6 +84,17 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+
+            schema:
+              $ref: '#/components/schemas/KYC_FillinRequest'
+            examples:
+              KYC_FillInBodyExample:
+                value:
+                  phoneNumber: '+34629255833'
```
AxelNennker, line 96: The content-type is application/json but the example is not JSON.
```suggestion
                  { "phoneNumber": "+34629255833" }

```

AxelNennker (changes requested):

AxelNennker:
> For this version of the KYC Fill-in API, the request body of the Fill-in Request does not include any parameters/attributes. The above mentioned AuthN/AuthZ/End User consent capturing processes can give the API provider/ MNO the 3rd party identity (with the Access Token) and the end user identity (with the ID Token), so the API provider/ MNO can identify the 3rd party / enterprise customer and the end user. Then, it is up to API provider/ MNO to decide which information the 3rd party will receive by calling the API, and also the API provider / MNO is responsible for security and privacy issues.

This PR MUST change the above text as well because with this PR the Fill-in Request does include a parameter.

There also should be text explaining what the semantic of providing the parameter is if the provided parameter does not match the phoneNumber associated with the acccess token.

GillesInnov35:
@AxelNennker , yes you're right, this must be removed.
Thanks
Gilles

GillesInnov35:
@AxelNennker , all. I've created PR https://github.com/camaraproject/KnowYourCustomer/pull/182 in order to update fill-in.md and to take into account Axel's comment.

could you have  look on it and if OK could you approve; and also this one. if OK also.
Thanks a lot
Gilles

PR #175, opened 2024-12-17
Updated Age Verification proposal aligned with Spring25

fernandopradocabrillo:
#### What type of PR is this?

* enhancement/feature

This PR include all the changes and suggestions that haven been made these past months regarding Age Verification in PR #50.
It is a clean version with errors and descriptions aligned with the latest version of Commonalities for Spring25 meta release.

Additionally, it contains a proposal to include two new fields in the response: `contentLock` and `parentalControl`. Both fields come from the feedback of our partners in the UK.

This PR is ment to replace PR #50 as it seems easier to work on a updated fresh copy.

Please leave any comments or changes you feel appropriate.

Thanks!

### DIFF
--- a/code/API_definitions/kyc-age-verification.yaml
+++ b/code/API_definitions/kyc-age-verification.yaml
@@ -11,7 +11,7 @@ info:
 
     The API defines one service endpoint:
 
-    - `POST /ageverify`
+    - `POST /verify`
 
       Takes the network subscription identifier (e.g. the mobile phone number for a mobile network subscriber) and checks if the age of the subscriber is older than the age threshold in the request.
       Additionally, as optional function, provides (1) if age check is done against another form of official identification or not (verifiedStatus), (2) an overall score of how certain is the response using both information provided in the request and information that the Operator holds(identityMatchScore), (3) an indicator on whether the subscription has any kind of content lock (contentLock) and (4) an indicator on whether the subscription has any kind of parental control activated (parentalControl).
@@ -23,7 +23,7 @@ info:
     - `phoneNumber`: The network subscription identifier (i.e. the phone number of the subscriber). [Required only in case a 2-legged flow has been agreed between API Provider and API Consumer following CAMARA directives; otherwise, phoneNumber must not be included]
     - `ageThreshold`: The age threshold from which the age of the user must be compared. [Required]
     - any of `idDocument`, `name`, `givenName`, `familyName`, `middleNames`, `familyNameAtBirth`, `birthdate`, `email`; additional subscriber's information to be checked in order to confirm that the subscriber is the contract's owner [Optional]
-    - `contentLock` and `parentLock`: These two flags allow the API Client to indicate that the corresponding response properties should be returned. [Optional]
+    - `includeContentLock` and `includeParentalControl`: These two flags allow the API Client to indicate that the corresponding response properties `contentLock` and `parentalControl` should be returned. Its implementation is optional for the API Provider, so in such cases, these parameters will be ignored. [Optional]
 
 
     ## Outputs
@@ -37,7 +37,7 @@ info:
     
     ## Errors
 
-    - If the authentication token is not valid, a `401 UNAUTHENTICATED` error is returned
+    - If the authentication access token is not valid, a `401 UNAUTHENTICATED` error is returned
     - If the API call contains a formatting or other any other syntactic error, a `400 INVALID_ARGUMENT` error is returned.
     - If the network subscription cannot be identified from the provided parameters (e.g. the subscription identifier is not associated with any customer of the CSP), a `404 IDENTIFIER_NOT_FOUND` error is returned.
     - If the API consumer has a valid access token but is not authorised to request an age verification for the specified network subscription, then a `403 PERMISSION_DENIED` error is returned.
@@ -97,9 +97,9 @@ paths:
       description: |
         Verify that the age of the subscriber associated with a phone number is equal to or greater than the specified age threshold value.
 
-        As it is possible that the person holding the contract and the person using the line may not be the same, the endpoint also admits a list of optional properties to be included in the request to improve the identification. The response may optionally include the `identityMatchScore` property with a value that indicates how certain it is that the information returned relates to the person that the API Client is requesting. To increase the reliability of the information returned, the operator may include in the response the `verifiedStatus` property, indicating whether the identity information in its possession has been verified against an official source.
+        As it is possible that the person holding the contract and the end-user of the subscription may not be the same, the endpoint also admits a list of optional properties to be included in the request to improve the identification. The response may optionally include the `identityMatchScore` property with a value that indicates how certain it is that the information returned relates to the person that the API Client is requesting. To increase the reliability of the information returned, the API Provider may include in the response the `verifiedStatus` property, indicating whether the identity information in its possession has been verified against an identification document legally accepted as an age verification document.
 
-        The response may also include `contentLock` and `parentalControl` properties to indicate if the line has any kind of content filtering enabled.
+        If the API Client indicates request properties `includeContentLock` or `includeParentalControl` with value `true` and the API Provider implements this functionality, then the response will also include `contentLock` and `parentalControl` properties to indicate if the subscription has any kind of content filtering enabled. On the other hand, if the request properties are not included or the API Client specifies value `false`, then the response properties will not be returned. If the API Provider doesn't implement this functionality, request properties will be ignored and response properties won't be returned in any case.
       operationId: verifyAge
       security:
         - openId:
@@ -125,8 +125,8 @@ paths:
                   familyNameAtBirth: YYYY
                   birthdate: '1978-08-22'
                   email: "federicaSanchez.Arjona@example.com"
-                  contentLock: true
-                  parentalControl: true
+                  includeContentLock: true
+                  includeParentalControl: true
               Three-Legged Access Token Example:
                 value:
                   ageThreshold: 18
@@ -138,8 +138,8 @@ paths:
                   familyNameAtBirth: YYYY
                   birthdate: '1978-08-22'
                   email: "federicaSanchez.Arjona@gmail.com"
-                  contentLock: true
-                  parentalControl: true
+                  includeContentLock: true
+                  includeParentalControl: true
 
       responses:
         '200':
@@ -208,50 +208,85 @@ components:
         - ageThreshold
       properties:
         ageThreshold:
-          type: "integer"
-          minimum: 0
-          maximum: 120
-          description: The age to be verified. The indicated range is a global definition of maximum and minimum values allowed to be requested. It is important to note that this range might be more restrictive in some implementations due to local regulations of a country i.e. A country does not allow to request for an age under 18. This limitation must be informed during the onboarding process.
+          $ref: "#/components/schemas/AgeThreshold"
         phoneNumber:
-          description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
-          type: string
-          pattern: '^\+[1-9][0-9]{4,14}$'
-          example: "+123456789"
+          $ref: "#/components/schemas/PhoneNumber"
         idDocument:
-          type: "string"
-          description: Id number associated to the official identity document in the country. It may contain alphanumeric characters.
+          $ref: "#/components/schemas/IdDocument"
         name:
-          type: "string"
-          description: Complete name of the customer, usually composed of first/given name and last/family/sur- name in a country.  Depending on the country, the order of first/give name and last/family/sur- name varies, and middle name could be included.  It can use givenName, middleNames, familyName and/or familyNameAtBirth. For example, in ESP, name+familyName; in NLD, it can be name+middleNames+familyName or name+middleNames+familyNameAtBirth, etc.
+          $ref: "#/components/schemas/Name"
         givenName:
-          type: "string"
-          description: First/given name or compound first/given name of the customer.
+          $ref: "#/components/schemas/GivenName"
         familyName:
-          type: "string"
-          description: Last name, family name, or surname of the customer.
+          $ref: "#/components/schemas/FamilyName"
         middleNames:
-          type: "string"
-          description: Middle name/s of the customer.
+          $ref: "#/components/schemas/MiddleNames"
         familyNameAtBirth:
-          type: "string"
-          description: Last/family/sur- name at birth of the customer.
+          $ref: "#/components/schemas/FamilyNameAtBirth"
         birthdate:
-          type: string
-          format: date
-          description: The birthdate of the customer, in RFC 3339 / ISO 8601 calendar date format (YYYY-MM-DD).
+          $ref: "#/components/schemas/Birthdate"
         email:
-          type: string
-          format: email
-          description: Email address of the customer in the RFC specified format (local-part@domain).
-        contentLock:
-          type: boolean
-          default: false
-          description: If this parameter is included in the request with value `true`, the response property `contentLock` will be returned. If it is not included or its value is `false`, the response property will not be returned.
-        parentalControl:
-          type: boolean
-          default: false
-          description: If this parameter is included in the request with value `true`, the response property `parentalControl` will be returned. If it is not included or its value is `false`, the response property will not be returned.
+          $ref: "#/components/schemas/Email"
+        includeContentLock:
+          $ref: "#/components/schemas/IncludeContentLock"
+        includeParentalControl:
+          $ref: "#/components/schemas/IncludeParentalControl"
 
+    AgeThreshold:
+      type: "integer"
+      minimum: 0
+      maximum: 120
+      description: The age to be verified. The indicated range is a global definition of maximum and minimum values allowed to be requested. It is important to note that this range might be more restrictive in some implementations due to local regulations of a country i.e. A country does not allow to request for an age under 18. This limitation must be informed during the onboarding process.
+
+    PhoneNumber:
+      description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
+      type: string
+      pattern: '^\+[1-9][0-9]{4,14}$'
+      example: "+123456789"
+
+    IdDocument:
+      type: "string"
+      description: Id number associated to the official identity document in the country. It may contain alphanumeric characters.
+
+    Name:
+      type: "string"
+      description: Complete name of the customer, usually composed of first/given name and last/family/sur- name in a country.  Depending on the country, the order of first/give name and last/family/sur- name varies, and middle name could be included.  It can use givenName, middleNames, familyName and/or familyNameAtBirth. For example, in ESP, name+familyName; in NLD, it can be name+middleNames+familyName or name+middleNames+familyNameAtBirth, etc.
+
+    GivenName:
+      type: "string"
+      description: First/given name or compound first/given name of the customer.
+
+    FamilyName:
+      type: "string"
+      description: Last name, family name, or surname of the customer.
+
+    MiddleNames:
+      type: "string"
+      description: Middle name/s of the customer.
+
+    FamilyNameAtBirth:
+      type: "string"
+      description: Last/family/sur- name at birth of the customer.
+
+    Birthdate:
+      type: string
+      format: date
+      description: The birthdate of the customer, in RFC 3339 / ISO 8601 calendar date format (YYYY-MM-DD).
+
+    Email:
+      type: string
+      format: email
+      description: Email address of the customer in the RFC specified format (local-part@domain).
+
+    IncludeContentLock:
+      type: boolean
+      default: false
+      description: If this parameter is included in the request with value `true`, the response property `contentLock` will be returned. If it is not included or its value is `false`, the response property will not be returned.
+
+    IncludeParentalControl:
+      type: boolean
+      default: false
+      description: If this parameter is included in the request with value `true`, the response property `parentalControl` will be returned. If it is not included or its value is `false`, the response property will not be returned.
 
     VerifyResponseBody:
       type: "object"
@@ -272,7 +307,7 @@ components:
 
     AgeCheck:
       type: "string"
-      description: Indicate `"true"` when the age of the user is the same age or older than the age threshold (age >= age threshold), and `"false"` if not (age < age threshold). If the Operator doesn't have enough information to perform the validation, a `not_available` can be returned. 
+      description: Indicate `"true"` when the age of the user is the same age or older than the age threshold (age >= age threshold), and `"false"` if not (age < age threshold). If the API Provider doesn't have enough information to perform the validation, a `not_available` can be returned. 
       enum:
         - 'true'
         - 'false'
@@ -280,13 +315,13 @@ components:
 
     VerifiedStatus:
       type: "boolean"
-      description: Indicates if the information provided was checked against another form of official identification held by the Operator.
+      description: Indicate `true` if the information provided has been compared against information based on an identification document legally accepted as an age verification document, otherwise indicate `false`.
 
     IdentityMatchScore:
       type: "integer"
       minimum: 0
       maximum: 100
-      description: The overall score of identity information available in the Operator, information either provided in the request body comparing it to the one that the Operator holds or directly using internal Operator's information. It is optional for the Operator to return the Identity match score.
+      description: The overall score of identity information available in the API Provider, information either provided in the request body comparing it to the one that the API Provider holds or directly using internal API Provider's information. It is optional for the API Provider to return the Identity match score.
 
     ContentLock:
       type: "string"
@@ -298,7 +333,7 @@ components:
     
     ParentalControl:
       type: "string"
-      description: Indicate `"true"` if the subscription associated with the phone number has any kind of parental control activated. and `"false"` if not. If the information is not available the value `not_available` can be returned.
+      description: Indicate `"true"` if the subscription associated with the phone number has any kind of parental control activated and `"false"` if not. If the information is not available the value `not_available` can be returned.
       enum:
         - 'true'
         - 'false'


code/API_definitions/age-verification.yaml
```diff
@@ -0,0 +1,586 @@
...
+                      - 503
+                  code:
+                    enum:
+                      - SERVICE_UNAVAILABLE
+          examples:
+            GENERIC_503_UNAVAILABLE:
+              description: Service is not available. Temporary situation usually related to maintenance process in the server side
+              value:
+                status: 503
+                code: SERVICE_UNAVAILABLE
```
eric-murray, line 559: There was a typo in CAMARA_common.yaml which should be corrected soon. See [here](https://github.com/camaraproject/Commonalities/pull/359).
```suggestion
                      - UNAVAILABLE
          examples:
            GENERIC_503_UNAVAILABLE:
              description: Service is not available. Temporary situation usually related to maintenance process in the server side
              value:
                status: 503
                code: UNAVAILABLE
```

code/API_definitions/age-verification.yaml
```diff
@@ -0,0 +1,586 @@
...
+        middleNames:
+          type: "string"
+          description: Middle name/s of the customer.
+        familyNameAtBirth:
+          type: "string"
+          description: Last/family/sur- name at birth of the customer.
+        birthdate:
+          type: string
+          format: date
+          description: The birthdate of the customer, in ISO 8601 calendar date format (YYYY-MM-DD).
```
eric-murray, line 225:
```suggestion
          description: The birthdate of the customer, in RFC 3339 / ISO 8601 calendar date format (YYYY-MM-DD).
```

code/API_definitions/age-verification.yaml
```diff
@@ -0,0 +1,586 @@
...
+                  ageThreshold: 18
+                  phoneNumber: '+34629255833' # Note: Only required in case a 2-legged flow has been agreed between API Provider and API Consumer; otherwise, do not include phoneNumber
+                  idDocument: 66666666q
+                  name: Federica Sanchez Arjona
+                  givenName: Federica
+                  familyName: Sanchez Arjona
+                  middleNames: Sanchez
+                  familyNameAtBirth: YYYY
+                  birthdate: '1978-08-22'                     
+                  email: "federicaSanchez.Arjona@gmail.com"
```
eric-murray, line 126:
```suggestion
              Two-Legged Access Token Example:
                value:
                  ageThreshold: 18
                  phoneNumber: '+34629255833' # Note: Only required in case a 2-legged flow has been agreed between API Provider and API Consumer; otherwise, do not include phoneNumber
                  idDocument: 66666666q
                  name: Federica Sanchez Arjona
                  givenName: Federica
                  familyName: Sanchez Arjona
                  middleNames: Sanchez
                  familyNameAtBirth: YYYY
                  birthdate: '1978-08-22'
                  email: "federicaSanchez.Arjona@gmail.com"
              Three-Legged Access Token Example:
                value:
                  ageThreshold: 18
                  idDocument: 66666666q
                  name: Federica Sanchez Arjona
                  givenName: Federica
                  familyName: Sanchez Arjona
                  middleNames: Sanchez
                  familyNameAtBirth: YYYY
                  birthdate: '1978-08-22'
                  email: "federicaSanchez.Arjona@gmail.com"
```

code/API_definitions/age-verification.yaml
fernandopradocabrillo: oh I wasn't aware, thanks @eric-murray !

eric-murray (approved):

eric-murray:
I'm happy with this, so approving. A couple of comments for you to think on:
- Linting is not enabled for this repository, so you might like to run that separately
- You might like to add some guidance around providing the `birthdate` information. If an API consumer really knew the date of birth of the customer, they wouldn't need to use this API! So I assume this information is useful only in that it allows us to verify the customer more quickly in case they API consumer knows the correct date of birth. But if it is incorrect, it will be ignored and other parameters used to verify the identity of the customer.

rartych:
@fernandopradocabrillo Thanks for this consolidated PR

The latest API Design guidelines add requirements for `api-name` and using it in file name and `servers url `.
I would also take the **location-verification API** as the reference.
Hence I suggest to set `api-name=kyc-age-verification` then we get:

- filename:  `kyc-age-verification.yaml`
- `url: "{apiRoot}/kyc-age-verification/vwip"`
- path: (verb)

 ```
paths:
  /verify:
```

-  security:

```
   - openId:
          - kyc-age-verification:verify
```
 - operationId - _OperationIds are defined in lowerCamelCase_
      `verifyAge` - in line with location verification

- tags - Title Case
         `Age Verification`

**Comment by fernandopradocabrillo** [comment]
> I'm happy with this, so approving. A couple of comments for you to think on:
> 
> * Linting is not enabled for this repository, so you might like to run that separately
> * You might like to add some guidance around providing the `birthdate` information. If an API consumer really knew the date of birth of the customer, they wouldn't need to use this API! So I assume this information is useful only in that it allows us to verify the customer more quickly in case they API consumer knows the correct date of birth. But if it is incorrect, it will be ignored and other parameters used to verify the identity of the customer.

@eric-murray Thank you for the review! I will pass the linter locally, and I'll raise a PR to include it in the repo later (after Holidays probably haha)

Regarding the birthdate, from TEF we agree... It was in the original proposal and I just left it there, but you are right, it doesn't make much sense.

wdyt about the birthdate @GillesInnov35 @ToshiWakayama-KDDI

**Comment by fernandopradocabrillo** [comment]
> @fernandopradocabrillo Thanks for this consolidated PR
> 
> The latest API Design guidelines add requirements for `api-name` and using it in file name and `servers url `. I would also take the **location-verification API** as the reference. Hence I suggest to set `api-name=kyc-age-verification` then we get:
> 
> * filename:  `kyc-age-verification.yaml`
> * `url: "{apiRoot}/kyc-age-verification/vwip"`
> * path: (verb)
> 
> ```
> paths:
>  /verify:
> ```
> 
> * security:
> 
> ```
>    - openId:
>           - kyc-age-verification:verify
> ```
> 
> * operationId - _OperationIds are defined in lowerCamelCase_
>   `verifyAge` - in line with location verification
> * tags - Title Case
>   `Age Verification`

Thanks @rartych, I have applied the changes to align with commonalities guidelines

**Comment by HuubAppelboom** [comment]
> I'm happy with this, so approving. A couple of comments for you to think on:
> 
> * Linting is not enabled for this repository, so you might like to run that separately
> * You might like to add some guidance around providing the `birthdate` information. If an API consumer really knew the date of birth of the customer, they wouldn't need to use this API! So I assume this information is useful only in that it allows us to verify the customer more quickly in case they API consumer knows the correct date of birth. But if it is incorrect, it will be ignored and other parameters used to verify the identity of the customer.

@eric-murray Hi Eric, this  birthdate information is what the use claims it to be. If it it does not match, you are sure the end user is not the contract owner, and you can not return a reasonable answer.  We propose to reply with "not_available" and not provide an identityMatchScore. See also the proposal I made here: https://github.com/camaraproject/KnowYourCustomer/issues/81#issuecomment-2425951384

**Comment by eric-murray** [comment]
@HuubAppelboom
> If it it does not match, you are sure the end user is not the contract owner, and you can not return a reasonable answer. 

OK. So the API also allows the API consumer to confirm whether the DoB they have been given by the customer is correct.

**Comment by HuubAppelboom** [comment]
@eric-murray Yes, the DoB is then used as a first tollgate check. If it matches, you continue to processing the other input parameters that have been provided with which you can calculate the aggregate identity match score.

**Comment by ToshiWakayama-KDDI** [comment]
> > @fernandopradocabrillo Thanks for this consolidated PR
> > The latest API Design guidelines add requirements for `api-name` and using it in file name and `servers url `. I would also take the **location-verification API** as the reference. Hence I suggest to set `api-name=kyc-age-verification` then we get:
> > 
> > * filename:  `kyc-age-verification.yaml`
> > * `url: "{apiRoot}/kyc-age-verification/vwip"`
> > * path: (verb)
> > 
> > ```
> > paths:
> >  /verify:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * security:
> > 
> > ```
> >    - openId:
> >           - kyc-age-verification:verify
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * operationId - _OperationIds are defined in lowerCamelCase_
> >   `verifyAge` - in line with location verification
> > * tags - Title Case
> >   `Age Verification`
> 
> Thanks @rartych, I have applied the changes to align with commonalities guidelines

Hi @Fernando, @Rafal, all,

Thank you for intensive review of Age Verification API.

But, from KDDI side, I cannot accept all of this kind of last minute fundamental changes.  We have already started implementation work, as this API has been discussed since January and the discussion had become stable before. 

From KDDI side, I would like to ask to revert url and path to the previous ones, i.e. 

- url: {apiRoot}/kyc-ageverify/vwip
- path: /ageverify
- Consecurtly I would propose to change the filename to: kyc-ageverify.yaml.

I do not think the above is not against commonalities guidelines.

Sorry for the late response due to the time difference, but your understanding would be appreciated.

Best reagrds,
Toshi
KDDI

**Comment by ToshiWakayama-KDDI** [comment]
Hi all,

I am checking other changes e.g. error responses, made by European colleagues during my night.  Please allow me to take some time and comments. 

Best regards,
Toshi
KDDI

**Comment by fernandopradocabrillo** [comment]
> Hi @fernando, @rafal, all,
> 
> Thank you for intensive review of Age Verification API.
> 
> But, from KDDI side, I cannot accept all of this kind of last minute fundamental changes. We have already started implementation work, as this API has been discussed since January and the discussion had become stable before.
> 
> From KDDI side, I would like to ask to revert url and path to the previous ones, i.e.
> 
> * url: {apiRoot}/kyc-ageverify/vwip
> * path: /ageverify
> * Consecurtly I would propose to change the filename to: kyc-ageverify.yaml.
> 
> I do not think the above is not against commonalities guidelines.
> 
> Sorry for the late response due to the time difference, but your understanding would be appreciated.
> 
> Best reagrds, Toshi KDDI

Hi @ToshiWakayama-KDDI,

Sorry, I think you mentioned this before. For me there is no problem in reverting to the previous name and aligning the filename.

I like this one better, by I can live with it 😄

**Comment by claraserranosolsona** [comment]
Hi @GillesInnov35 , @ToshiWakayama-KDDI , are you happy with the PR? 
Let us know if you have any additional comment, as I understood the objective would be to close the RC within this week 
Many thanks in advance
Clara

**Comment by eric-murray** [comment]
@ToshiWakayama-KDDI 
> But, from KDDI side, I cannot accept all of this kind of last minute fundamental changes. We have already started implementation work, as this API has been discussed since January and the discussion had become stable before.

The registered name for this API in the [API Backlog WG](https://github.com/camaraproject/APIBacklog/blob/main/documentation/APIbacklog.md) is, and always has been, "KYC Age Verification". If you want it to be known as "KYC Age Verify", then you should submit a PR to that WG to get it changed. Even then, the name in the URL should be `/kyc-age-verify` (i.e. `url: {apiRoot}/kyc-age-verify/vwip`) following Commonalities rules.

File: code/API_definitions/kyc-age-verification.yaml [hunk 4]
```diff
@@ -0,0 +1,598 @@
...
+
+  version: wip
+  contact:
+    email: sp-kyc@lists.camaraproject.org
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: "{apiRoot}/kyc-ageverify/vwip"
```
eric-murray, line 80:
```suggestion
  - url: "{apiRoot}/kyc-age-verify/vwip"
```

GillesInnov35:
hi @claraserranosolsona, I've discussed internally with my team. Regarding parentControl and contentBlock, they submitted those ideas
- each country should have possibility to decide/manage if those kind of information should be returned 
- the requester should indicate clearly if information are expected (scope or claims).
I've scheduled a point this afternoon. I guess I'll have a decision from our side 
Thanks a lot
BR
Gilles

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+      description: |
+        Verify that the age of the subscriber associated with a phone number is equal to or greater than the specified age threshold value.
+
+        As it is possible that the person holding the contract and the person using the line may not be the same, the endpoint also admits a list of optional properties to be included in the request to improve the identification. The response may optionally include the `identityMatchScore` property with a value that indicates how certain it is that the information returned relates to the person that the API Client is requesting. To increase the reliability of the information returned, the operator may include in the response the `verifiedStatus` property, indicating whether the identity information in its possession has been verified against an official source.
+
+        The response may also include `contentLock` and `parentalControl` properties to indicate if the line has any kind of content filtering enabled.
+      operationId: verifyAge
+      security:
+        - openId:
+          - kyc-ageverify:ageverify
```
rartych, line 105: If we keep the `\verify` in paths then the scope should be:
```suggestion
          - kyc-age-verify:verify
```

code/API_definitions/kyc-age-verification.yaml
rartych: The full request would look like `"{apiRoot}/kyc-age-verify/vwip/verify"`, but we can decide and change the names in separate PR as it is not related to API functionality.

The guidelines were update with: https://github.com/camaraproject/Commonalities/pull/333

ToshiWakayama-KDDI: Hi @eric-murray , @rartych ,

Thanks.  Then  I will submit a PR to API Backlog WG.  If the API name is KYC Ageverify, it would be url "{apiRoot}/kyc-ageverify/vwip".  Am I correct?

Thanks,
Toshi

eric-murray: > Thanks. Then I will submit a PR to API Backlog WG. If the API name is KYC Ageverify, it would be url "{apiRoot}/kyc-ageverify/vwip". Am I correct?

Yes, though expect some debate on calling an API "KYC Ageverify".

code/API_definitions/kyc-age-verification.yaml
ToshiWakayama-KDDI: Hi @rartych ,

Thanks.  I have asked to keep the original path 'ageverify', instead of 'verify'.  If url becomes either of kyc-ageverify or kyc-age-verify, 'ageverify' can be used, I understand.  Correct?

Thanks,
Toshi

GillesInnov35:
@claraserranosolsona, after an discussion at Orange see bellow our quick comments 
- those information are not determinist while the other are determinist (ageCheck, verifiedStatus). 
- if we decide to include those information, they should be returned only if the developer ask for them (request body or claims) 
- some countries will probalbly decide not to return those information. In that case, will the returned value be unavailable ? 
Thanks 
BR
Gilles

code/API_definitions/kyc-age-verification.yaml
rartych: There is no clear guideline but in section 3.1 there is such a sentence:

> When the POST method is used, the resource in the path must be a verb (e.g. retrieve-location and not location) to differentiate from an actual resource creation.

One can say that the verb is `verify`, so the kebab case should be used: `verify-age` or `age-verify` (but it starts with not a verb)

`"{apiRoot}/kyc-ageverify/vwip/ageverify"` looks redundant to me, but we can leave it for final review

ToshiWakayama-KDDI:
> Hi @GillesInnov35 , @ToshiWakayama-KDDI , are you happy with the PR? Let us know if you have any additional comment, as I understood the objective would be to close the RC within this week Many thanks in advance Clara

Hi @claraserranosolsona , 

I should comment it is not possible for us to complete checking of these big bunches of changes from the previous version.  I am still discussing with my implementation / product teams.  Fortunately, I do not think we have to close the RC within this week now, and I think we should extend the target date of the RC.

Regarding the addtional paramters, 'contentLock' and 'parentalControl', as our quick response, we should be fine with them as long as they are optional and they don't add mandatory logic to the API backend.  (In addition, I am not sure if they are related to verifying age, but it is fine with me.)

Many thanks,
Toshi

code/API_definitions/kyc-age-verification.yaml
ToshiWakayama-KDDI: Hi @rartych , @eric-murray , @fernandopradocabrillo ,

I had one question during my internal discussion regarding this part.

> Security:
>   -openId:
>       -kyc-ageverify:ageverify

(Currently the scope is kyc-ageverify:ageverify.)

Can we keep the first part 'kyc-ageverify', even though we change the filename to kyc-age-verification.yaml and change the url to {apiRoot}/kyc-age-verification/vwip ?  In another wey, should the first part 'kyc-ageverify' be aligned with 'api-name', i.e. kyc-age-verification?

Many thanks,
Toshi

ToshiWakayama-KDDI:
Hi @fernandopradocabrillo , @GillesInnov35 , @eric-murray , @rartych , all

I have been working internally on the recently proposed changes to get the Age Verificaiton yaml aligned with the latest Commonalities.  We are investigating some work around, in order for us to accept the proposed changes.  Now we are on new-year holiday in Japan, so, it is to be confirmed early next week, i.e. on 6th January at the earliest.  Please give me some more time.

Let me double-check the proposed changes as below:
 
- API-name: KYC Age Verification (decided by API Backlog and GSMA Product WS from the beginning)


-> Based on the API name, 

- filename: kyc-age-verification.yaml

- url: "{apiRoot}/kyc-age-verification/vwip"

->As 'paths' should be verb, 

- paths:  
       /verify:

->based on the above, 

- security: 
     -openId:  
    -kyc-age-verification:verify


Please let me know if my understanding above is not correct.

Many thanks,
Toshi

ToshiWakayama-KDDI:
Hi @fernandopradocabrillo , @GillesInnov35 , @eric-murray , @rartych , all

I am glad to inform you that we/KDDI can accept the proposed changes as written down in my previous comment, above, for Age Verification initial version / Spring25 Meta version.

So, no API name change will be challenged in API Backlog.

I would like to ask Fernando to modify the yaml file accordingly.

Thanks.

code/API_definitions/kyc-age-verification.yaml
ToshiWakayama-KDDI: As commented in the main conversation, url "{apiRoot}/kyc-age-verification/vwip" is now fine with us/KDDI.

Thanks
Toshi

code/API_definitions/kyc-age-verification.yaml
ToshiWakayama-KDDI: As commented in the main conversation, "kyc-age-verification:verify" is now fine with us/KDDI.

Thanks,
Toshi

AxelNennker:
Note: Please remember the OIDF presentation from March 2024 which included Age Verification
https://github.com/camaraproject/KnowYourCustomer/blob/f6c75cf4ed2db6d08287d23dfe468a034f8ebfdd/documentation/MeetingMinutes/MOM-2024-03-19.md#oidf-presentation-re-age-verification

This OIDF specification defines that Advanced Syntax: https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#name-example-age-verification

Interesting side-note: Requests should be signed

> Integrity protection and authentication of authentication requests can be achieved in particular by
using Pushed Authorization Requests [[RFC9126](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9126)] to send requests server-to-server with authentication of the RP, or
using JWT-Secured Authorization Requests [[RFC9101](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9101)] to sign the authentication request parameters.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+    - `verifiedStatus`: true if the information provided was checked against another form of official identification, otherwise false.
+    - `identityMatchScore`: The overall score of identity information available in the Operator, information either provided in the request body comparing it to the one that the MNO holds or directly using internal MNO's information. It is optional for the Operator to return the Identity match score.
+    - `contentLock`: Indicates if the subscription associated with the phone number has any kind of content lock (i.e certain web content blocked).
+    - `parentalControl`: Indicates if the subscription associated with the phone number has any kind of parental control activated.
+    
+    ## Errors
+
+    - If the authentication token is not valid, a `401 UNAUTHENTICATED` error is returned
+    - If the API call contains a formatting or other any other syntactic error, a `400 INVALID_ARGUMENT` error is returned.
+    - If the network subscription cannot be identified from the provided parameters (e.g. the subscription identifier is not associated with any customer of the CSP), a `404 IDENTIFIER_NOT_FOUND` error is returned.
```
AxelNennker, line 41: What is an "authentication token"? API access token? Isn't that error then 401 Unauthorized? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

IDENTIFIER_NOT_FOUND can be used to implement a Telco Finder. If the API consumer has a valid 3-legged access token then the this error cannot happen, right? Is this error for 2-legged tokens only?

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+    - If the API consumer has a valid access token but is not authorised to request an age verification for the specified network subscription, then a `403 PERMISSION_DENIED` error is returned.
```
AxelNennker, line 42: How can that be? The API consumer should never get a valid access token in this case, right?
At onboarding time of the API consumer includes the scope "kyc-age-verification:verify" or not. And then the access token is created or not.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+paths:
+  /verify:
+    post:
+      tags:
+        - Age Verify
+      summary: Verify Age Threshold
+      description: |
+        Verify that the age of the subscriber associated with a phone number is equal to or greater than the specified age threshold value.
+
+        As it is possible that the person holding the contract and the person using the line may not be the same, the endpoint also admits a list of optional properties to be included in the request to improve the identification. The response may optionally include the `identityMatchScore` property with a value that indicates how certain it is that the information returned relates to the person that the API Client is requesting. To increase the reliability of the information returned, the operator may include in the response the `verifiedStatus` property, indicating whether the identity information in its possession has been verified against an official source.
```
AxelNennker, line 99: How does including additional data improve the "identification"? What is the threat model?

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+        - $ref: '#/components/parameters/x-correlator'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VerifyRequestBody'
+            examples:
+              Two-Legged Access Token Example:
+                value:
```
AxelNennker, line 116: The content has the content-type application/json but the example value is no JSON.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+                  ageThreshold: 18
+                  phoneNumber: '+34629255833' # Note: Only required in case a 2-legged flow has been agreed between API Provider and API Consumer; otherwise, do not include phoneNumber
+                  idDocument: 66666666q
+                  name: Federica Sanchez Arjona
+                  givenName: Federica
+                  familyName: Sanchez Arjona
+                  middleNames: Sanchez
+                  familyNameAtBirth: YYYY
+                  birthdate: '1978-08-22'
+                  email: "federicaSanchez.Arjona@gmail.com"
```
AxelNennker, line 126:
```suggestion
                  email: "federicaSanchez.Arjona@example.com"
```

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+                value:
+                  ageThreshold: 18
+                  idDocument: 66666666q
+                  name: Federica Sanchez Arjona
+                  givenName: Federica
+                  familyName: Sanchez Arjona
+                  middleNames: Sanchez
+                  familyNameAtBirth: YYYY
+                  birthdate: '1978-08-22'
+                  email: "federicaSanchez.Arjona@gmail.com"
```
AxelNennker, line 137:
```suggestion
                  email: "federicaSanchez.Arjona@example.com"
```

AxelNennker (changes requested):
Are the additional functions essential for the commercial success? Otherwise they lead to more trouble than they are worth. Keep it simple.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
+openapi: 3.0.3
+info:
+  title: Know Your Customer Age Verification
+
+  description: |
+
+    # Summary
+    This API provides the customer with the ability to check if the user of the line is older than a provided age, in order to provide API customer's age-restricted services, access to its age-restricted website etc..
```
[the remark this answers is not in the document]
HuubAppelboom: I think the goal to verify the age of the end user in most use cases, not that of the subscriber (although the subscriber data is used for the verification)

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+
+    # API functionality
+
+    The API defines one service endpoint:
+
+    - `POST /ageverify`
+
+      Takes the network subscription identifier (e.g. the mobile phone number for a mobile network subscriber) and checks if the age of the subscriber is older than the age threshold in the request.
+      Additionally, as optional function, provides (1) if age check is done against another form of official identification or not (verifiedStatus), (2) an overall score of how certain is the response using both information provided in the request and information that the Operator holds(identityMatchScore), (3) an indicator on whether the subscription has any kind of content lock (contentLock) and (4) an indicator on whether the subscription has any kind of parental control activated (parentalControl).
```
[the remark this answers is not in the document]
HuubAppelboom: Official Identification can be defined as Identity Document which is legally accepted as a means to verify someone's identity. We don't want to define which trust framework is used for this, but more whether data has been verified or not at all.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+    - `phoneNumber`: The network subscription identifier (i.e. the phone number of the subscriber). [Required only in case a 2-legged flow has been agreed between API Provider and API Consumer following CAMARA directives; otherwise, phoneNumber must not be included]
+    - `ageThreshold`: The age threshold from which the age of the user must be compared. [Required]
+    - any of `idDocument`, `name`, `givenName`, `familyName`, `middleNames`, `familyNameAtBirth`, `birthdate`, `email`; additional subscriber's information to be checked in order to confirm that the subscriber is the contract's owner [Optional]
+
+
+    ## Outputs
+
+    If successful, a JSON object is returned containing the following data:
+    - `ageCheck`: Indicate 'true' when the age of the user is the same age or older than the age threshold (age >= age threshold), and 'false' if not (age < age threshold).  Otherwise, 'not_available'.
+    - `verifiedStatus`: true if the information provided was checked against another form of official identification, otherwise false.
```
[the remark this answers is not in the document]
HuubAppelboom: note that government issued is not always correct, for example for driver licenses this can be different, like that it is issued by semi-government organisations. Suggest to change it to legally accepted identifcation document. What is legally accepted can also vary per country.

code/API_definitions/kyc-age-verification.yaml
```diff
@@ -0,0 +1,598 @@
...
+    - `identityMatchScore`: The overall score of identity information available in the Operator, information either provided in the request body comparing it to the one that the MNO holds or directly using internal MNO's information. It is optional for the Operator to return the Identity match score.
```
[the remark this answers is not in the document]
HuubAppelboom: The MNO can publish the algorithm used, if need be.  Note that algorithm may vary depending on the quality of the data the MNO holds, so it is difficult to come with a algorithm which will work in all cases.

code/API_definitions/kyc-age-verification.yaml
HuubAppelboom: What needs to be verified is whether the end user is the same person as the subscriber (user) to a certain level of degree of certainty. That is what calculating the identity match score is for

HuubAppelboom:
> Note: Please remember the OIDF presentation from March 2024 which included Age Verification https://github.com/camaraproject/KnowYourCustomer/blob/f6c75cf4ed2db6d08287d23dfe468a034f8ebfdd/documentation/MeetingMinutes/MOM-2024-03-19.md#oidf-presentation-re-age-verification
> 
> This OIDF specification defines that Advanced Syntax: https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#name-example-age-verification
> 
> Interesting side-note: Requests should be signed
> 
> > Integrity protection and authentication of authentication requests can be achieved in particular by
> > using Pushed Authorization Requests [[RFC9126](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9126)] to send requests server-to-server with authentication of the RP, or
> > using JWT-Secured Authorization Requests [[RFC9101](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9101)] to sign the authentication request parameters.

@AxelNennker If you look at the Advanced Syntax specification, I think this specification is much to broad on the one hand for age verification (you can define a lot more complex use cases) which also ha  a strong risk in it that it will be used for cases which are not intended. I think we are much better of with a well defined API, including wel defined options, than giving relying parties a toolbox by which can do a lot of things.  Furthermore, if you look into the specs, I doubt you can for example also  calculate a Jaro-Winkler score with it

code/API_definitions/kyc-age-verification.yaml
HuubAppelboom: This is probably for the 2 legged flow, not?

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
AxelNennker: @HuubAppelboom I am struggling with the word "line". I amended my suggestion and reintroduced "user" but removed "of the line".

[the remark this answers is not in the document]
HuubAppelboom: @AxelNennker It would be more precise to use the word End-User, as it is defined in the Glossary of Identity and Consent Management

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
AxelNennker: I updated my suggestion

code/API_definitions/kyc-age-verification.yaml
AxelNennker: To clarify: my comment is about this item on the list
> - If the API consumer has a valid access token but is not authorised to request an age verification for the specified network subscription, then a `403 PERMISSION_DENIED` error is returned.

With 2-legged flow there is a scope in the client credentials token request, right?
https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2

The 2-legged token request with OAuth2 client_credentials might look like in this example, right?:
```
POST /token HTTP/1.1
Host: server.example.com
Authorization: Bearer eyJhbGciOiJFUzI1NiJ9.ewogICJqdGkiOiJteUpXVElkMDAxIiwKICAic3ViIjoiMzgxNzQ2MjM3NjIiLAogICJpc3MiOiIzODE3NDYyMzc2MiIsCiAgImF1ZCI6Imh0dHA6Ly9sb2NhbGhvc3Q6NDAwMC9hcGkvYXV0aC90b2tlbi9kaXJlY3QvMjQ1MjMxMzgyMDUiLAogICJleHAiOjE1MzYxNjU1NDAsCiAgImlhdCI6MTUzNjEzMjcwOAp9Cg.YB4gdhWUGRjWEsEbKDs7-G2WFH2oYz7bAEP5AtegHXInkY9ncA2V3IoA6O_HVQuFxyCRIklrxsMk32MfNF_ABA
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials&scope="dpv:FraudPreventionAndDetection kyc-age-verification:verify"
```

The resulting access token is then valid for the scope `kyc-age-verification:verify`. So, `403 PERMISSION_DENIED` cannot happen, I think.

code/API_definitions/kyc-age-verification.yaml
AxelNennker: What about this scenario: A minor is using their parent's mobile phone to access age-restricted content on the API consumer's website.

The minor enters all the parents attributes into the API consumer webpage. 
Variant 1) CIBA
The API consumer sends a signed CIBA request to the API provider. The API Provider sends a message to the subscriber's mobile phone that is in the hands of the minor. The minor enters e.g. a received OTP code into the API consumers webpage and the API consumer retrieves the access token.
Then the API consumer requests age-verification with those parameter:

```
POST /ageverify
Host: camaraapi.example.com
Authorization: Bearer SplxlOBeZQQYbYS6WxSbIA
Content-Type: application/json

{
  "ageThreshold": 18,
  "idDocument": "66666666q",
  "name": "Federica Sanchez Arjona",
  "givenName": "Federica",
  "familyName": "Sanchez Arjona",
  "middleNames": "Sanchez",
  "familyNameAtBirth": "YYYY",
  "birthdate": "1978-08-22",
  "email": "federicaSanchez.Arjona@gmail.com"
}
```

Sending the attributes in the age-verification process does help in verifying whether the end user (minor) is the same person as the subscriber (user)(parent).

Variant 2) OIDC authorization code flow
The API consumer sends a signed OIDC code request with `max_age=0`.
Note `max_age` is [mandatory to implement for all OIDC server](https://openid.net/specs/openid-connect-core-1_0.html#ServerMTI).
The minor does not know the subscriber credentials and the flow stops. Access to the age-protected contents is prevented.

Security considerations: 
- all requests for code and tokens for age-verification MUST be signed by the API consumer. The API provider MUST validate the signatur. 
- all access tokens SHOULD be sender-constrained e.g. by using DPoP.

AxelNennker:
> > Note: Please remember the OIDF presentation from March 2024 which included Age Verification https://github.com/camaraproject/KnowYourCustomer/blob/f6c75cf4ed2db6d08287d23dfe468a034f8ebfdd/documentation/MeetingMinutes/MOM-2024-03-19.md#oidf-presentation-re-age-verification
> > This OIDF specification defines that Advanced Syntax: https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#name-example-age-verification
> > Interesting side-note: Requests should be signed
> > > Integrity protection and authentication of authentication requests can be achieved in particular by
> > > using Pushed Authorization Requests [[RFC9126](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9126)] to send requests server-to-server with authentication of the RP, or
> > > using JWT-Secured Authorization Requests [[RFC9101](https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#RFC9101)] to sign the authentication request parameters.
> 
> @AxelNennker If you look at the Advanced Syntax specification, I think this specification is much to broad on the one hand for age verification (you can define a lot more complex use cases) which also has a strong risk in it that it will be used for cases which are not intended. I think we are much better of with a well defined API, including wel defined options, than giving relying parties a toolbox by which can do a lot of things. Furthermore, if you look into the specs, I doubt you can for example also calculate a Jaro-Winkler score with it, so on the other hand the toolbox is also too limited.

@HuubAppelboom I agree, I just wanted to remind the WG that OIDF is discussing age-verfication for ages. Age verification was already discussed in OpenId 1.0 and early [IIW](https://internetidentityworkshop.com/)s. Currently most work is done in the field of Verifiable Credentials/Presentations, I think.

AxelNennker:
There are examples of law that requires that e.g. in a web forum participants MUST be children, that is UNDER a certain age. 

E.g.: [COPPA](https://www.ftc.gov/legal-library/browse/rules/childrens-online-privacy-protection-rule-coppa) and [CIPA](https://www.fcc.gov/consumers/guides/childrens-internet-protection-act)

Maybe change ageThreshHold to ageLowerBound and introduce ageUpperBound, or something?

AxelNennker:
Should the API provider prevent the API consumer from testing all values from 0 to 100 to find the birth-year of a user?

The API consumer can send many requests for one user and if we are not careful the API provider can find the exact birth-year.

AxelNennker:
Is `max_age=0` mandatory for age-verification?
I think, that IF we want the age of the End-User to be verified then the End-User MUST authenticate.
Otherwise, all we verify is the age of the subscriber, I think.

code/API_definitions/kyc-age-verification.yaml
HuubAppelboom: @AxelNennker  Regarding variant 1. If the end user (a minor) is entering all the details of the susbcriber (a parent) with the API Consumer, the Age Verification will of course not work.  We assume here that the minor is entering his/her own data, and email address etc. In case the minor is clever and knows about this and enters all the data, the API Consumer can always send a confirmation mail that he is granted access to the service, see what happens....

If you are looking for a age verifcation which is done at a higher level of assurance, you are probably better of with a different product, like for instance an eIDAS wallet.

AxelNennker:
What would be the `purpose` usually associated with this API?
https://w3c.github.io/dpv/2.0/dpv/modules/purposes.html#AgeVerification seems very generic.

If the API consumer is a bank and the legal age for a minor having a bank account is 14, what `purpose` would the API consumer then send? This is not a problem of this subproject, but somewhere this has to be defined, I think.

HuubAppelboom:
> There are examples of law that requires that e.g. in a web forum participants MUST be children, that is UNDER a certain age.
> 
> E.g.: [COPPA](https://www.ftc.gov/legal-library/browse/rules/childrens-online-privacy-protection-rule-coppa) and [CIPA](https://www.fcc.gov/consumers/guides/childrens-internet-protection-act)
> 
> Maybe change ageThreshHold to ageLowerBound and introduce ageUpperBound, or something?

I would leave it for now as is. I have not seen any internet service yet where adults are excluded.  Usually it is something like an internet filter which excludes content for adults, and you can only be excluded from the filter if you prove you are an adult.

HuubAppelboom:
> What would be the `purpose` usually associated with this API? https://w3c.github.io/dpv/2.0/dpv/modules/purposes.html#AgeVerification seems very generic.
> 
> If the API consumer is a bank and the legal age for a minor having a bank account is 14, what `purpose` would the API consumer then send? This is not a problem of this subproject, but somewhere this has to be defined, I think.

@AxelNennker I think you can still ctegorize this under fraud prevention (the use is actually not allowed access to a service, and is committing fraude to ge the access by claiming he is older than 18)

HuubAppelboom:
> Is `max_age=0` mandatory for age-verification? I think, that IF we want the age of the End-User to be verified then the End-User MUST authenticate. Otherwise, all we verify is the age of the subscriber, I think.

@AxelNennker We see a lot of use cases where the API consumer wants to have a bit more assurance than asking whether you are older than 18.  And in many cases, the authentication is not necessary. For example for a subscrption on a monthly lottery ticket you need to be older than 18.  What happens in that minors do register the tickets on their own name (otherwise you run the risk of not getting paid in case of a big prize), but lie about their age.  Authentication is in this case not necessary, just confirming that the name is associated with a phone number and a date of birth which is 18+ is sufficient.  The API consumer is always looking for as low as possible friction, and it is best to leave the decision on what authentication is need to the API Consumer.

code/API_definitions/kyc-age-verification.yaml
HuubAppelboom: In case the end user is excluded/blocked from API services this may come about with a 2 legged flow.  You may want to offer this kind of blocking as a service for example for VIPs or other special interest groups, where privacy concerns may prevalate.

HuubAppelboom:
> Should the API provider prevent the API consumer from testing all values from 0 to 100 to find the birth-year of a user?
> 
> The API consumer can send many requests for one user and if we are not careful the API provider can find the exact birth-year.

@AxelNennker The same applies to other attributes, like for example housenumber, or first name.  Best is probably to monitor for this kind of behaviour, and lock accounts that show abuse or search behaviour.

fernandopradocabrillo:
Hi, @GillesInnov35, @ToshiWakayama-KDDI, @HuubAppelboom 
As discussed in the last meeting with @claraserranosolsona, I have updated the proposal including the properties `contentLock`and `parentLock` as optional input flags in the request body to indicate if the corresponding response properties should be returned.
Thanks!

GillesInnov35 (commented):
hi @fernandopradocabrillo , concerning contentLock an parentControl, name of flag in the request should not be the same as it is 2 different parameters (a boolean and a string). WDYT
Gilles

HuubAppelboom:
> Is `max_age=0` mandatory for age-verification? I think, that IF we want the age of the End-User to be verified then the End-User MUST authenticate. Otherwise, all we verify is the age of the subscriber, I think.

For many cases, legitimate interests can be used, and in this case you don't need to authenticate at all, as long as you keep it a verification product, and the level of assurance that can ben obtained (including the optional identityMatchScore is sufficient for the API consumer.

In case a higher level of assurance regarding age is required or legitimate interest can not be used (depending on the use case), you may need another product, which is more similar to a eIDAS wallet (or at least something of a similar level of assurance).
In that case, and alternative approach to max_age=0 could be to solve it in the consent acquistion process of CAMARA, and make sure that next to acquiring the explict end user consent (with a tick box) you also let the subcriber provide the credentials required.  This is however a different product in my opnion, similar to teh difference between KYC Match and KYC Fill-in.

fernandopradocabrillo:
> hi @fernandopradocabrillo , concerning contentLock an parentControl, name of flag in the request should not be the same as it is 2 different parameters (a boolean and a string). WDYT Gilles

I also thought about it yerterday while doing it, but I ended up going with the same names. We can go for something like: `includeParentalControl` and `includeContentLock`. Or in the response change to `parentalControlOn` or `parentalControlActive`.

I don't have any preference

GillesInnov35:
hi @fernandopradocabrillo , thanks a lot for your proposal (includeParentalControl and includeContentLock), it suits me well. The objective is to allow consumer to request those 2 added parameters or not.

All, WDYT ?

Gilles

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
fernandopradocabrillo: I think we can include this, can we rephrase it a bit? I find it a little bit confusing, but I get the idea

```suggestion
    - `verifiedStatus`: Indicate `true` if the information provided has been compared against information based on an identification document legally accepted as an age verification document, otherwise indicate `false`.
```

fernandopradocabrillo:
> Hi @fernandopradocabrillo , @GillesInnov35 , @eric-murray , @rartych , all
> 
> I am glad to inform you that we/KDDI can accept the proposed changes as written down in my previous comment, above, for Age Verification initial version / Spring25 Meta version.
> 
> So, no API name change will be challenged in API Backlog.
> 
> I would like to ask Fernando to modify the yaml file accordingly.
> 
> Thanks.

Hi @ToshiWakayama-KDDI, It has been modified accordingly! Thank you very much

code/API_definitions/kyc-age-verification.yaml
fernandopradocabrillo: Yes, it is the access token, I'll update the text to reflect this more clear.

Regarding the IDENTIFIER_NOT_FOUND not found error, it is the error accepted in latest Commonalities release. It's common for all APIs that go to meta Spring25.

fernandopradocabrillo:
Thanks @ToshiWakayama-KDDI, @GillesInnov35  I have updated the name proposal and the descriptions to reflect the use of this functionality, please take a look and let me know if it suits you.

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
AxelNennker: I changed the suggestion to "End-User"

AxelNennker (changes requested):

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
AxelNennker: I think attributes of a person that are not well defined, will get us into trouble and will lead to interoperability problems. What are parental control exactly? If the end-user is of legal age, e.g. age_over_18 and age_over_21, BUT still there are parental controls in place, then could that mean the end-user is mentally-challenged and some if their rights were lawfully revoked? "parental control" would then be health data, which has special protection in most jurisdictions. Similar with contenLock? 
 I think these additional attribute, parental control or content block or ..., need to be specified. 
 I would add them to fill-in and match, because I these attributes should not be mixed with age-verification.

AxelNennker (changes requested):

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
claraserranosolsona: Hi Axel,

parentalControl or contentLock are capabilities offered in some MNOs to lock content for specific lines.
These parameters give indication that, even if the owner of the contract is an adult (it will always be), the line may not be in use by him, but by other family members. So directly related with age verification API.
These parameters have been already validated by markets with age verification commercial interest and have been introduced as optional so, in case an MNO or country can not expose them, it is not mandatory.

Let us know if you have any other question.

Regards,
Clara

code/API_definitions/kyc-age-verification.yaml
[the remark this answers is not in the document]
AxelNennker: @fernandopradocabrillo update to your suggested text

code/API_definitions/kyc-age-verification.yaml
AxelNennker: @HuubAppelboom Then I guess I am looking for text that replaces "to improve the identification".

HuubAppelboom: @AxelNennker 

Here is a a suggestion:  replace "to improve the identification" with "to verifiy whether the End-User is the same person as the Subscriber"  
or alternatively replave it with  "to verify whether the person using the line is the same person as the one holding the contract"

ToshiWakayama-KDDI (approved):
LGTM.  Thanks very much.

GillesInnov35 (approved):

GillesInnov35:
hi all, for includeContentLock or includeParentalControl I've made a proposition. I think that 'not_available' may be be returned in case of the MNO do not implement functionality. WDYT 
Thanks
Gilles

code/API_definitions/kyc-age-verification.yaml
ToshiWakayama-KDDI: Thanks @GillesInnov35 ,

But, from our side, as I commented before, we would like `contentLock` and `parentalControl` to be purely optional fundamentally (because of the very late proposal), meaning that API Providers/Operators do not have to implement any logic related to these two parameters and that `includeContentLock` and/or `includeParentalControl` can be ignored if a API Provider/Operator has not been implemented.  So, the current definition is fine.

If 'not_available' should be returned, API Providers/Operators have to implement required logic to understand the two attributes and to return 'not_available'.  It may be better to descript somewhere that if `includeContentLock` and/or `includeParentalControl` is put into the request and `ContentLock` and/or `ParentalControl` is not returned, it means the API Provider/Operator does not have the related function.  But, anyway, from our side, the current definition is fine, and your understanding would be appreciated.

Many thanks,
Toshi

GillesInnov35: Hi @ToshiWakayama-KDDI , thanks, I understand your comment. 
What I mean is as far as the 2 attributes are part of the design, the API Provider should manage the attribute's response even if the functionality has not been developed. If the consumer input them in the request body best practices may require a response. 

Gilles

PR #177, opened 2025-01-07
Add KYC-Fill-In examples for consent collection, access token requests and API use

AxelNennker:
#### What type of PR is this?
* documentation

This PR shows:
- how user consent can be requested using OIDC authorization code flow
- the use of kyc-fill-in:set-all when requesting an OIDC code
- explains that the subject identifier for the API is associated with the access token
- explains that the scopes for the API is associated with the access token

No extra phoneNumber request parameter is needed when using 3-legged / front-end flow.
Using an extra phoneNumber parameter in the API request should be restricted to 2-legged flow.

### DIFF
--- a/documentation/API_documentation/KYC-Fill-in.md
+++ b/documentation/API_documentation/KYC-Fill-in.md
@@ -38,4 +38,142 @@ The API provides the following endpoint:
 
 * An endpoint to request information related to an end user against the acount data bound to their phone number.
 
+## Examples
 
+As always examples are non-normative.
+
+### Get user consent for a set of scopes and purpose
+
+Example OIDC authorization code flow request requesting consent for all user-attributes and purpose dpv:FraudPreventionAndDetection.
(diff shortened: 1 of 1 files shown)


ToshiWakayama-KDDI:
Hi @AxelNennker ,

Thanks.  It seems to me that the added parts are explaining OIDC Authorisation flow and examples, so, there is no reason to object it.  However, I have not seen this kind of OIDC flow explanation in other APIs, so, it may be better to discuss and approve first in Commonalities or ICM, not KYC.  Concerned a bit.

WDYT, all?

Best regards,
Toshi

GillesInnov35:
hi @AxelNennker , @ToshiWakayama-KDDI , I think. that what is added in the yaml is more a Getting Started which show be exposed on the exposition gateway than a specifications.
I don't think it should be added in my opinion. 

Gilles

ToshiWakayama-KDDI:
Thanks, @GillesInnov35 .  I agree with you.  So, we can close this PR, can't we?  Or, if we need furthre discussion at least, we can conclude no impact to KYC Fill-in Spring25 Meta.

Thanks,
Toshi

PR #178, opened 2025-01-07
fix example response value to match content-type application/json

AxelNennker:
#### What type of PR is this?

* correction

The example provided has a content-type of application/json but the content was not JSON.

### DIFF
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -99,27 +99,25 @@ paths:
               examples:
                 KYC_Fillin200Example:
                   value:
-                    phoneNumber: '+34629255833'
-                    idDocument: 66666666q
-                    name: Federica Sanchez Arjona
-                    givenName: Federica
-                    familyName: Sanchez Arjona
-                    nameKanaHankaku: federica
-                    nameKanaZenkaku: Ｆｅｄｅｒｉｃａ
-                    middleNames: Sanchez
-                    familyNameAtBirth: YYYY
-                    address: Tokyo-to Chiyoda-ku Iidabashi 3-10-10
-                    streetName: Nicolas Salmeron
-                    streetNumber: 4
-                    postalCode: 1028460
-                    region: Tokyo
-                    locality: ZZZZ
-                    country: JP
-                    houseNumberExtension: 36
-                    birthdate: '1978-08-22'
-                    email: abc@example.com
-                    gender: male
-
+                    {
+                        "phoneNumber": "+34629255833",
+                        "idDocument": "66666666q",
+                        "name": "Federica Sanchez Arjona",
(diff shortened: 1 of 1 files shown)


fernandopradocabrillo:
Hi @AxelNennker, I think this is not needed in OAS3. The response and request examples follow the json-schema standard and it is assumed by default that the format is json when the content is application/json

AxelNennker:
My reading of the below yaml snippet is that the the content-type is application/json which the example value fails to be.

```yaml
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/KYC_FillinResponse'
              examples:
                KYC_Fillin200Example:
                  value:
                    phoneNumber: '+34629255833'
                    idDocument: 66666666q
                    name: Federica Sanchez Arjona
                    givenName: Federica
                    familyName: Sanchez Arjona
                    nameKanaHankaku: federica
                    nameKanaZenkaku: Ｆｅｄｅｒｉｃａ
                    middleNames: Sanchez
                    familyNameAtBirth: YYYY
                    address: Tokyo-to Chiyoda-ku Iidabashi 3-10-10
                    streetName: Nicolas Salmeron
                    streetNumber: 4
                    postalCode: 1028460
                    region: Tokyo
                    locality: ZZZZ
                    country: JP
                    houseNumberExtension: 36
                    birthdate: '1978-08-22'
                    email: abc@example.com
                    gender: male

```
Please have a look here at example for "example:"
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md

```
requestBody:
  content:
    'application/json':
      schema:
        $ref: '#/components/schemas/Address'
      examples:
        foo:
          summary: A foo example
          value: {"foo": "bar"}
        bar:
          summary: A bar example
          value: {"bar": "baz"}
    'application/xml':
      examples:
        xmlExample:
          summary: This is an example in XML
          externalValue: 'http://example.org/examples/address-example.xml'
    'text/plain':
      examples:
        textExample:
          summary: This is a text example
          externalValue: 'http://foo.bar/examples/address-example.txt'
```

fernandopradocabrillo:
@AxelNennker Take a look at all CAMARA APIs and the official documentation for OAS3: https://swagger.io/docs/specification/v3_0/adding-examples/

fernandopradocabrillo:
To add a bit more info, in visual ui tools this is also represented correctly with actual format:

image
image

the format is json by default, there is no need to include the { } characters as this is also a yaml file using a json-schema representation

ToshiWakayama-KDDI:
Thanks, @fernandopradocabrillo , @AxelNennker .
So, we can close this PR without merging, can't we?  Or, if we need furthre discussion, at least we can conclude no impact to KYC Fill-in Spring25 Meta.

Thanks,
Toshi

AxelNennker:
You are correct. Closing this.

PR #184, opened 2025-01-17
Update kyc-fill-in.yaml

### LINKED ISSUE #183: Align kyc-fill-in with commonalities for meta spring25
**Problem description**
Align kyc-fill-in with commonalities for meta spring25

**Possible evolution**
Align kyc-fill-in with commonalities for meta spring25 as it has been done for kyc-tenure and kyc-match

GillesInnov35:
align kyc-fill-in with Commonalities Spring25

#### What type of PR is this?

* enhancement/feature

align kyc-fill-in with Commonalities Spring25

Fixes #183

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/kyc-fill-in.yaml
+++ b/code/API_definitions/kyc-fill-in.yaml
@@ -26,11 +26,10 @@ info:
 
     * When calling this API, at the beginning, there should be required processes for Authentication / Authorisation / End User Consent capturing.  As those processes are to be defined as CAMARA commonality standards, they are out of scope of this API definition, however, use of the OpenID Connect (OIDC) is stated as security scheme.  **As an important note**, capturing end user consent is necessary, because this API provides end user information (PII).
 
-    * For this version of the KYC Fill-in API, the request body of the Fill-in Request does not include any parameters/attributes.  The above mentioned AuthN/AuthZ/End User consent capturing processes can give the API provider/ MNO the 3rd party identity (with the Access Token) and the end user identity (with the ID Token), so the API provider/ MNO can identify the 3rd party / enterprise customer and the end user.  Then, it is up to API provider/ MNO to decide which information the 3rd party will receive by calling the API, and also the API provider / MNO is responsible for security and privacy issues.
+    * The above mentioned AuthN/AuthZ/End User consent capturing processes can give the API provider/ MNO the 3rd party identity (with the Access Token) and the end user identity (with the ID Token), so the API provider/ MNO can identify the 3rd party / enterprise customer and the end user.  Then, it is up to API provider/ MNO to decide which information the 3rd party will receive by calling the API, and also the API provider / MNO is responsible for security and privacy issues.
 
     * For example, below is a potential operation:
       * when making contract with API provider/ MNO, a 3rd party / enterprise customer receives its 3rd party identity and also decide which information (attributes) it will receive for its API call
-      * when calling this API, the client of a 3rd party / enterprise customer put its 3rd party identity (only) in the request body
       * then, API provider / MNO will provide information (attributes) which the 3rd party / enterprise customer is allowed to receive by the contract.
 
 
@@ -40,7 +39,6 @@ info:
 
     * An endpoint to request information related to an end user against the account data bound to their phone number.
 
-
     ## Authorization and authentication
 
     The "Camara Security and Interoperability Profile" provides details on how a client requests an access token. Please refer to Identify and Consent Management (https://github.com/camaraproject/IdentityAndConsentManagement/) for the released version of the Profile.
@@ -49,15 +47,26 @@ info:
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
+    # Identifying the phoneNumber from the access token
+
+    This API requires the API consumer to identify a phoneNumber as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional phoneNumber field, which therefore MUST be provided.
+    - When a three-legged access token is used, the subject will be uniquely identified from the access token. So, the optional phoneNumber is not used for identifying the subject, but, if it is provided in the API request body, it needs to match with the one associated with the access token.
+
+    ## Error handling:
+
+    - If the subject cannot be identified from the access token and the optional phoneNumber field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code. This is an error typically, but not uniquely, returned in 2-legged scenarios when the access token doesn't identify a single subscription.
+    - If there is a mismatch between the provided phone number in the request body and the phone number associated with the access token, the server will return an error with the `403 INVALID_TOKEN_CONTEXT` error code.
+
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 
-  version: 0.2.0
+  version: wip
   x-camara-commonalities: 0.4.0
 
 servers:
-  - url: '{apiRoot}/kyc-fill-in/v0.2'
+  - url: '{apiRoot}/kyc-fill-in/vwip'
 
     variables:
       apiRoot:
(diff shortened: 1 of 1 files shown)


ToshiWakayama-KDDI:
Hi @GillesInnov35 ,

Thanks very much for this.  May I ask you for clarification.  I understand your proposal to add the optional phoneNumber input to KYC Fill-in is for 2-legged access flow cases.  Is my understanding correct?  Or, is your proposal for 3-legged access flow cases, as well?

Thanks, 
Toshi

GillesInnov35:
hi @ToshiWakayama-KDDI , yes PR https://github.com/camaraproject/KnowYourCustomer/pull/174 concerns the addition phone number for 2-Legged only, in my understanding 3-Legged is not concerned.
Thanks
Gilles

ToshiWakayama-KDDI:
Thanks, @GillesInnov35 ,

In that case, as for Line 50 ~ 54 for 'Identifying the phoneNumber from the access token' and Line 56 ~ 59 for 'Error handling', we can use the standard text provided as Commonalities v0.5 template, can't we?  We don't need to make it aligned with that of KYC Match, I think.

What I called the standard text provided as Commonalities v0.5 template above is:

# Identifying the phoneNumber from the access token

This API requires the API consumer to identify a phone number as the subject of the API as follows:

- When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` field, which therefore MUST be provided.

- When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.


## Error handling

- If the subject cannot be identified from the access token and the optional phoneNumber field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.

 - If the subject can be identified from the access token and the optional `phoneNumber` field is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same phone number is identified by these two methods, as the server is unable to make this comparison.

===

WDYT?

Thanks,
Toshi

code/API_definitions/kyc-fill-in.yaml
```diff
@@ -142,28 +165,43 @@ components:
...
+          example: "+123456789"
+          
     KYC_FillinResponse:
       type: object
       properties:
 
         phoneNumber:
           type: string
-          description: The phone number of the customer stored on the Operator's system.  A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
+          description: TA public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
```
ToshiWakayama-KDDI, line 202: Hi @GillesInnov35 ,

Sorry, a quick question, please.  What is 'TA public identifier'?

Thanks,
Toshi

code/API_definitions/kyc-fill-in.yaml
```diff
@@ -49,15 +47,26 @@ info:
...
+    - If the subject cannot be identified from the access token and the optional phoneNumber field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code. This is an error typically, but not uniquely, returned in 2-legged scenarios when the access token doesn't identify a single subscription.
+    - If there is a mismatch between the provided phone number in the request body and the phone number associated with the access token, the server will return an error with the `403 INVALID_TOKEN_CONTEXT` error code.
+
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 
-  version: 0.2.0
+  version: wip
   x-camara-commonalities: 0.4.0
```
ToshiWakayama-KDDI, line 66: Thanks, @GillesInnov35 ,

A quick suggestion to change 0.4.0 to 0.5.0.  (Or, will this be done by Pre-release PR?)

Thanks,
Toshi

fernandopradocabrillo: it can be done here

```suggestion
  x-camara-commonalities: 0.5.0

```

code/API_definitions/kyc-fill-in.yaml
```diff
@@ -49,15 +47,26 @@ info:
...
+
+    This API requires the API consumer to identify a phoneNumber as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional phoneNumber field, which therefore MUST be provided.
+    - When a three-legged access token is used, the subject will be uniquely identified from the access token. So, the optional phoneNumber is not used for identifying the subject, but, if it is provided in the API request body, it needs to match with the one associated with the access token.
+
+    ## Error handling:
+
+    - If the subject cannot be identified from the access token and the optional phoneNumber field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code. This is an error typically, but not uniquely, returned in 2-legged scenarios when the access token doesn't identify a single subscription.
+    - If there is a mismatch between the provided phone number in the request body and the phone number associated with the access token, the server will return an error with the `403 INVALID_TOKEN_CONTEXT` error code.
```
fernandopradocabrillo, line 59: This is the behavior in kyc-match but I don't know if you have defined the same here. This error is not included in the API, you have the 422 UNNECESSARY_IDENTIFIER. So you have 2 options:
1. Include the 403 INVALID_TOKEN_CONTEXT and allow to send the phoneNumber in the body in 3-legged scenarios like we have in kyc-match. In this case the implementator will have to implement the verifications between body and token.
2. Leave the 422 UNNECESSARY_IDENTIFIER like in the rest of CAMARA APIs and not allow to send the phoneNumber in the body in 3-legged scenarios

ToshiWakayama-KDDI:
Thanks, @GillesInnov35 , 

You have removed 422 UNNECESSARY_IDENTIFIER error, so you took Option 1, correct?  

In that case, 403_INVALID_TOKEN_CONTEXT should be included.  Could you add something like the below into Line 373 - 384 around?

                    enum:
                      - INVALID_TOKEN_CONTEXT

            GENERIC_403_INVALID_TOKEN_CONTEXT:
              description: Reflect some inconsistency between information in some field of the API and the related OAuth2 Token
              value:
                status: 403
                code: INVALID_TOKEN_CONTEXT
                message: "phoneNumber is not consistent with access token."


Many thanks,
Toshi

GillesInnov35:
hi @ToshiWakayama-KDDI , that'ok . It has been done.
Thanks
Gilles

GillesInnov35:
so finally, according to the previous meeting this mormning, we should be aligned with Commonalities v5.0 for kyc-fill-in as it has been done for Tenure, is it OK for you ?
so now we need to 
- add error code  **"UNNECESSARY_IDENTIFIER"**    
- remove **INVALID_TOKEN_CONTEXT** 


``

    Generic422:
      description: Unprocessable Content
      headers:
        x-correlator:
          $ref: "#/components/headers/X-Correlator"
      content:
        application/json:
          schema:
            allOf:
              - $ref: "#/components/schemas/ErrorInfo"
              - type: object
                properties:
                  status:
                    enum:
                      - 422
                  code:
                    enum:
                      - SERVICE_NOT_APPLICABLE
                      - MISSING_IDENTIFIER
                      - UNNECESSARY_IDENTIFIER


    Generic403:
      description: Forbidden
      headers:
        x-correlator:
          $ref: "#/components/headers/X-Correlator"
      content:
        application/json:
          schema:
            allOf:
              - $ref: "#/components/schemas/ErrorInfo"
              - type: object
                properties:
                  status:
                    enum:
                      - 403
                  code:
                    enum:
                      - PERMISSION_DENIED
                      - 
``



and modify the description with the bellow text

    # Identifying the phone number from the access token

    This API requires the API consumer to identify a phone number as the subject of the API as follows:
    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` field, which therefore MUST be provided.

    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.

    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.

    ## Error handling:
    - If the subject cannot be identified from the access token and the optional `phoneNumber` field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.

    - If the subject can be identified from the access token and the optional `phoneNumber` field is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same phone number is identified by these two methods, as the server is unable to make this comparison.



 Then it'll be aligned with Tenure errors specifications.
Could you validate or update  to close the PR

Thanks a lot 
Gilles

ToshiWakayama-KDDI (approved):
Thank you so much.  LGTM.

PR #181, opened 2025-01-16
Pre release r2.1

fernandopradocabrillo:
#### What type of PR is this?

Add one of the following kinds:
* subproject management

#### What this PR does / why we need it:
Includes documentation for preparing the pre-release r2.1 for KYC API Family

#### Special notes for reviewers:
Test plan is being provided in the following PR: https://github.com/camaraproject/KnowYourCustomer/pull/180

PR #186, opened 2025-02-06
Remove non-mandatory errors

fernandopradocabrillo:
#### What type of PR is this?

* cleanup

As recommended by latest Commonalities, we can remove all common errors that don't have an specific purpose for this APIs, such as 429, 500, 502, 503, 504, etc

This has been also mention by RM in #181


GillesInnov35:
thanks @fernandopradocabrillo, it should also be applied for tenure from what I understand.

GillesInnov35 (approved):
LGTM thanks

ToshiWakayama-KDDI (approved):
Thank you so much.  LGTM.  Sorry for the delay.

PR #208, opened 2025-06-10
docs: clarify that match = true means perfect match with score 100

### LINKED ISSUE #198: [KYC-Match]: Having a match with `true` should be defined as a perfect match with a score of `100`
### Discussed in https://github.com/camaraproject/KnowYourCustomer/discussions/191

<div type='discussions-op-text'>

<sup>Originally posted by **maxl2287** April 11, 2025</sup>
When I want to implement this API, what should be the expectation: 

- `true` = 100 score, as it is written that it matches the value from the operator system ?

From my PoV this should be the way, as otherwise it would not be a "match".

What is your opinion about that?

Thanks in advance.
</div>

maxl2287:
#### What type of PR is this?

* enhancement/feature
* documentation
* tests

- Added detailed descriptions to MatchResult and MatchScoreResult fields in the OpenAPI spec.
- Specified that matchScore is only returned when match is 'false' with values between 0 and 99.
- Confirmed that match = true implies a perfect match and corresponds to a score of 100, with no score returned in that case.

Fixes #198


GillesInnov35 (approved):
LGTM

maxl2287:
@ToshiWakayama-KDDI yes I will do so asap :)

PR #217, opened 2025-07-18
[KYC Match]: Minor correction. Removing idDocumentMatchScore from test definition.

### LINKED ISSUE #216: when idDocument does not match idDcumentMatchScore validation require or not. clairifcation needed
**Problem description**

**when idDocument does not match then idDocumentMatchScore validation is  not required** .  As per BDD test we must validate idDocumentMatchSCore in api response.

https://github.com/camaraproject/KnowYourCustomer/blob/r2.2/code/Test_definitions/kyc-match.feature#L83

In .YAML we dont see any  field idDocumentMatchScore.
https://github.com/camaraproject/KnowYourCustomer/blob/r2.2/code/API_definitions/kyc-match.yaml#L174

**Expected action**

idDocumentMarchScore  validation required when iddocument does not match??

**Additional context**

trehman-gsma:
#### What type of PR is this?

* cleanup

Minor correction. Removing `idDocumentMatchScore` as an example response_score_property_path. The field does not exist in the main KYC API spec.

<!-- Automatically closes linked issue when PR is merged.
Fixes #216

#### Changelog input

```
Update kyc-match.feature to remove an example value
```

### DIFF
--- a/code/Test_definitions/kyc-match.feature
+++ b/code/Test_definitions/kyc-match.feature
@@ -80,7 +80,6 @@ Feature: CAMARA Know Your Customer Match API, v0.3.0 - Operation KYC_Match
 
         Examples:
             | request_property_path     | response_property_path        | response_score_property_path  |
-            | $.idDocument              | $.idDocumentMatch             | $.idDocumentMatchScore        |
             | $.name                    | $.nameMatch                   | $.nameMatchScore              |
             | $.givenName               | $.givenNameMatch              | $.givenNameMatchScore         |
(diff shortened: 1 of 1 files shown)


fernandopradocabrillo (approved):
Hi @trehman-gsma, I'll leave the approve for this one in case you want to proceed, but in the new repo we merged a PR to fix several other issues on the test plan, including other properties that should not be present as they don't exist in the PAI, same as idDocumentMatchScore.

Take a look in case you want to remove all of them here: https://github.com/camaraproject/KnowYourCustomerMatch/pull/26.

I think there is no need to fix the rest of the things or align the test.

